PR #9, opened 2016-01-06
Fix player.cpp crash when m_pActiveItem = null

IgnacioFDM:
It used to check if m_pActiveItem was != 0 after referencing it the line before.

(Fucked up last PR)


IgnacioFDM:
I don't know if this should be REGAMEDLL_FIXES, because I believe the original valve code has got this right (at least I don't get this crash with valve's mp.dll)

s1lentq:
it's my typo
thank you for the PR

IgnacioFDM:
Awesome work by the way!

PR #13, opened 2016-01-26
Add fine grained infinite round settings

IgnacioFDM:
Sometimes you need to only block one check, such as round time.


s1lentq:
i think it would be better to read "round_infinite.string" once and convert to bits or enums.

IgnacioFDM:
There you go. Don't mind github's stupid indentation issue.

PR #26, opened 2016-02-16
SV_StudioSetupBones fix for non-players

jonatan1024:


s1lentq:
@jonatan1024 please can you explain why do it?
if for non-players UTIL_GetPlayerGaitYaw will always return 0, right?

jonatan1024:
Yes, but it will reset the entity's yaw. Every non-player entity has yaw set to zero because of this bug.

PR #27, opened 2016-03-02
End round command

IgnacioFDM:
Went the command way because the cvar way (like sv_restart) seems pretty counter intuitive IMO (actions should be commands, and CVARS configurations), but feel free to edit whatever you want.


s1lentq:
please, you can merge your two commits into one?

IgnacioFDM:
ok

PR #40, opened 2016-06-11
Fix crash when custom SV_StudioSetupBones exported to the engine

a1batross:
This fixes crash, in case ReGameDLL_CS running with a local client. 

If local client do a tracing for studio models, it calls SV_StudioSetuoBones with a NULL pEdict. So it must be checked before we call GetPlayerYaw/Pitch to prevent crash.

### DIFF
--- a/regamedll/dlls/animation.cpp
+++ b/regamedll/dlls/animation.cpp
@@ -1062,16 +1062,8 @@ void SV_StudioSetupBones(model_t *pModel, float frame, int sequence, const vec_t
 		/*static */float q3[MAXSTUDIOBONES][4], q4[MAXSTUDIOBONES][4];
 
 		float_precision s, t;
-		if( pEdict )
-		{
-			s = GetPlayerYaw(pEdict);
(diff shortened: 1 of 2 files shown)


WPMGPRoSToTeMa:
In original gamedll this check is inside GetPlayerYaw and GetPlayerPitch functions.

a1batross:
I see that VARS(), CBaseEntity::Instance() have checks for null pointers, but regamedll anyway crashes while tracing from client. 

Crash causes on Linux(GCC 5.2.0) and Android(GCC 4.9.x) on Xash3D Engine. Additional check fixes it. 

I had not checked it on GoldSource Linux yet.

WPMGPRoSToTeMa:
Original GetPlayerYaw/GetPlayerPitch stops before dynamic_cast.

a1batross:
So I moved check to a new place.

s1lentq:
@a1batross thank you for your report, really missing check for null.

PR #94, opened 2017-01-06
Optimize conditional toggle to bitwise operation

In-line:


WPMGPRoSToTeMa:
Wow, short and beautiful. Maybe add macro `ToggleFlag`?

In-line:
```
#define SetBits(a,b) ( (a) |= (b) )
#define ClearBits(a,b) ( (a) &= ~(b) )
#define IsSetBits(a,b) ( (a) & (b) )
#define InvertBits(a,b) ( (a) ^= (b) )
#define ToggleBits InvertBits
#define IsNotSetBits(a,b) ( ( ~(a) )  & (b) )
```
Why not use something like this?

s1lentq:
@In-line it doesn't violate original logic of the code, you can do without wrapping the macro.

In-line:
@s1lentq i'm just a little macro fun.

In-line:
@s1lentq Are they another places in code where we can do the same trick ? I tried to search but unable to find.

s1lentq:
@In-line [1](https://github.com/s1lentq/ReGameDLL_CS/blob/9b955f2/regamedll/dlls/pathcorner.cpp#L64-L72), [2](https://github.com/s1lentq/ReGameDLL_CS/blob/9b955f2/regamedll/dlls/pathcorner.cpp#L76-L84).

Issue #107, opened 2017-02-10
abs UB on integral types (negate UB) and float->int UB


WPMGPRoSToTeMa:
@spark512 have found some bug: the server crashes with emit sound error (bad volume) on deathrun_clouds_v2 when you activating the first trap. This trap activates func_breakable which has `9999999999999999999` HP. So, the problem is located in [CBreakable::Die](https://github.com/s1lentq/ReGameDLL_CS/blob/7c9d59101b67525a35b0b3a31e17159ab5d42fbd/regamedll/dlls/func_break.cpp#L679):
```C++
fvol = RANDOM_FLOAT(0.85f, 1.0f) + (Q_abs(int(pev->health)) / 100.0f);
```
There we have 2 kinds of UB:

- float->int conversion. If the value cannot fit into the destination type, the behavior is undefined. When converting `9999999999999999999` from float to int I get `-2147483648` on my platform.
- Taking absolute value from INT_MIN (if INT_MIN < -INT_MAX). The behavior is undefined if the result cannot be represented by the return type (negate operation has the same problem). When taking absolute value for `-2147483648` I get `-2147483648` on my platform.

Then we get negative `fvol` and crash...

So, I don't know about other places in code, but this UB should be checked and fixed in all the code. This not only refers to ReGameDLL, but to ReHLDS and other projects.

spark512:
`Deathrun: Restarting. Reason: Fatal error - EMIT_SOUND: volume = 1073741824`

So that number is basically a garbage value due to improper float -> int conversion..

WPMGPRoSToTeMa:
@spark512 nope, float values are pushed into another stack, so it prints fully garbage, not converted float.

spark512:
Any progress on this?

Can we just clamp the value taken from pev->health? That will prevent the overflow and will prevent a negative value being passed.

Also, why is the health being transformed to an int? We are working with floats, fvol is a float..

IgnacioFDM:
It's probably best to clamp pev->health of map entities on spawn. I might be wrong (haven't really checked), but there may be some HP to int casts client side or in netcode too, as happens with player health.

WPMGPRoSToTeMa:
@IgnacioFDM entity health is not transmitted to client.

spark512:
pev->health is being retrieved here, and it's a float. Why not just check abs on float? There are plenty of code snippets for that.

Or we can just get it ourselves if we check pev_health > 0 || pev_health < 0..

PR #126, opened 2017-03-06
in regards to issue #107

spark512:


spark512:
@IgnacioFDM What about float abs instead of regular abs for integer? There is no need to convert the int to float just to check abs, and then back to float...

In-line:
@spark512 You need to consult a profiler, before optimizations like this

IgnacioFDM:
Yes, I meant fabs. It will be much faster because there is no branching, and it's done in a single FABS instruction (there is an SSE one too I believe). Also it's shorter code, and self explanatory.

spark512:
There is no need for that conversion to int, Q_abs is overloaded and can work with floats directly. Thanks @WPMGPRoSToTeMa.

theAsmodai:
Abs for float faster than for integers.

WPMGPRoSToTeMa:
[Original](https://github.com/ValveSoftware/halflife/blob/5d761709a31ce1e71488f2668321de05f791b405/dlls/func_break.cpp#L599) code is:
```C++
fvol = RANDOM_FLOAT(0.85, 1.0) + (abs(pev->health) / 100.0);
```
But it is compiled with C `abs`, not C++ `abs` which is overloaded for `float`. I think VALVE developers forgot to compile code with C++ `abs`. So we can remove `int` conversions in similar `abs` calls.

s1lentq:
@spark512 Please wrap your the fixed code in the preprocessor REGAMEDLL_FIXES, otherwise it will fail.

spark512:
@s1lentq done

PR #141, opened 2017-04-29
Break loop in OLD_CheckBuyZone, OLD_CheckBombTarget, OLD_CheckRescueZone

justgo97:
stop loop when match is found


In-line:
1. It's better to add `REHLDS_FIXES` check, because behavior changed.
2. Make PR title more informative

justgo97:
1. done
2. any ideas ?

PR #142, opened 2017-04-30
Added SPEC team to autojoin

justgo97:


s1lentq:
could you also please add some details about it  [here](https://github.com/s1lentq/ReGameDLL_CS/blob/a5dac3c267d13e60c64b1c6704611871b9027a53/dist/game.cfg#L91).
something like **any/SPEC/CT/T**

PR #143, opened 2017-05-01
Added mp_disable_buyzone cvar

justgo97:
usefull for many gamemods jailbreak/zombie/deathmatch, mp_disable_buyzone 1/0 by default its set to 0 can be enabled/disabled in-game.


justgo97:
just noticed this xD

```
// -1 - means no time limit
21	//  0 - disable buy
22	//
23	// Default value: "1.5"
24	mp_buytime 0.25
```
mp_buytime "0"

PR #144, opened 2017-05-01
dead see alive messages

justgo97:


justgo97:
added by mistake, sorry

regamedll/dlls/client.cpp
```diff
@@ -756,7 +756,7 @@ void Host_Say(edict_t *pEntity, BOOL teamonly)
 		if (teamonly && client->m_iTeam != player->m_iTeam)
 			continue;
 
-		if ((client->pev->deadflag != DEAD_NO && !bSenderDead) || (client->pev->deadflag == DEAD_NO && bSenderDead))
```
WPMGPRoSToTeMa: It can be shorter `(client->pev->deadflag == DEAD_NO) == bSenderDead`.

justgo97: (client->pev->deadflag == DEAD_NO && bSenderDead) no difference ?

justgo97: (client->pev->deadflag == DEAD_NO) == bSenderDead
oh i see
1#: seems fine
(client->pev->deadflag == DEAD_NO) = 1 -> player alive
bSenderDead = 1 -> sender is dead
1 = 1 -> continue
#2: but what about here ?
(client->pev->deadflag == DEAD_NO) = 0 -> player dead
bSenderDead = 0 -> sender is alive
1 = 1 -> continue

WPMGPRoSToTeMa: No.
`(client->pev->deadflag != DEAD_NO && !bSenderDead) || (client->pev->deadflag == DEAD_NO && bSenderDead)`
or
`(client->pev->deadflag == DEAD_NO) == bSenderDead`
means
`(isReceiverAlive && !isSenderAlive) || (!isReceiverAlive && isSenderAlive)`
or
`isReceiverAlive != isSenderAlive` in shorter form. So, we skip this iteration if receiver and sender not both alive or not both dead.

`client->pev->deadflag == DEAD_NO && bSenderDead` means that receiver should be alive and sender should be dead.

justgo97: yeah thats what i wanted dead to see alive, i guess you wanted to make it shorter and keep default hlds behavor ?

WPMGPRoSToTeMa: > yeah thats what i wanted dead to see alive

Dead to see alive, but not alive to see dead? If so, then you're right.

> i guess you wanted to make it shorter and keep default hlds behavor ?

Yup, i just made it shorter.

justgo97: > Dead to see alive, but not alive to see dead?

yeah exactly

>  But you need to make it configurable via some new cvar and REGAMEDLL_ADD.

yeah i will thats why i closed the pull request once i m done adding the cvar i will reopen it.

PR #146, opened 2017-05-02
Added mp_chat_flags cvar

justgo97:
mp_chat_flags 0 = disable (default)
flags: 
a = dead see alive chat (only team)
b = alive see dead chat (only team)
.
.
more flags soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ There are several software requirements for building Regamedll_CS:
 </ol>
 
 ### Checking requirements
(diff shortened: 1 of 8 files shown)


s1lentq:
please, you can testing via gradle your changes before committing.
too many builds are failed.

justgo97:
i have linux and it all tests gets passed and in windows when the gradle doesn't work for me.

PR #148, opened 2017-05-04
completly disable buyzone when mp_buytime is 0

justgo97:
currently when you set mp_buytime 0 it shows buyicon and its allows you to open weapons menu and when you try buy somthing it show "0 seconds passed since..." now with this patch no icon will be shown and menu will not open.

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -6375,8 +6375,14 @@ void CBasePlayer::HandleSignals()
 {
 	if (CSGameRules()->IsMultiplayer())
 	{
-		if (!CSGameRules()->m_bMapHasBuyZone)
-			OLD_CheckBuyZone(this);
+		
+#ifdef REGAMEDLL_ADD
+		if (buytime.value != 0.0f)
+#endif
+		{			
+			if (!CSGameRules()->m_bMapHasBuyZone)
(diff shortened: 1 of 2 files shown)


Phantomsq:
I think, these changes must be in REGAMEDLL_FIXES

justgo97:
> I think, these changes must be in REGAMEDLL_FIXES define

its not a bug older hlds didn't have this option.

Phantomsq:
Isn't it logically that if we will set 0 to this cvar, buyzone must be disabled without showing buyicon and message:

> "0 seconds passed since..."

I still think, that it's FIXES

justgo97:
old gamedll doesn't allow value less then 0.25 with regamedll there is two new options -1 = unlimited time and 0 = disabled. so its new features more then a fix or mayby fix for a features, REGAMEDLL_FIXES should be used only for a bug that was in engine and it got fixed which is not the case here.

Phantomsq:
Dunno, anyway imo this is fix

justgo97:
yes you can call it a fix but its not the fix that should be under REGAMEDLL_FIXES.

Phantomsq:
Ok

justgo97:
> Ok, as you want

its not as i want it how it i think it should be.

PR #157, opened 2017-06-03
Add more sv_alltalk modes

In-line:
### New modes:

2. The teammates hear each other (dead hear alive vise versa)
3. Same as 2, but spectators hear everybody.

This PR will get good use in Clan War servers.
Idea was suggested by N3UR0.


shelru:
Why you think that we need additional cvar values? Maybe just make an API, everyone can do what he needs.

In-line:
There is already API for CanPlayerHearPlayer, but it's convenient to have them without third-party modules/plugins. Because of missing feature like this players use TeamSpeak or another soft to communicate with each other. I guess cw servers are pretty​ common.

justgo97:
isnt better to use flags like for example "a" = "hear dead", "b" = "hear alive", "c" "hear spec" ect.. so you can combine two flags "abc"

In-line:
@justgo97 How we can specify teams like that ? Or situation when dead should hear alive, but not vice versa?

justgo97:
a flag for each team/situation ?

IgnacioFDM:
I think the flag solution is overcomplicated.

PR #159, opened 2017-06-05
Optimize CBasePlayer::SendWeatherInfo

In-line:
Return immediately, when we found something, instead of calling `UTIL_FindEntityByClassname` 2 times.

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -6740,27 +6740,28 @@ void CBasePlayer::SendHostageIcons()
 
 void CBasePlayer::SendWeatherInfo()
 {
-	CBaseEntity *pPoint = UTIL_FindEntityByClassname(NULL, "env_rain");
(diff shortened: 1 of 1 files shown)


regamedll/dlls/player.cpp
```diff
@@ -6740,27 +6740,28 @@ void CBasePlayer::SendHostageIcons()
...
-	CBaseEntity *pPoint = UTIL_FindEntityByClassname(NULL, "env_rain");
-	CBaseEntity *pPoint2 = UTIL_FindEntityByClassname(NULL, "func_rain");
+	CBaseEntity *pEnt;
 
-	if (pPoint || pPoint2)
+	auto SendReceiveW = [&](BYTE byte)
 	{
 		MESSAGE_BEGIN(MSG_ONE, gmsgReceiveW, NULL, pev);
-			WRITE_BYTE(1);		// rainy weather
+				WRITE_BYTE(byte);		// snowy weather
```
souvikdas95: fix the comment.

PR #172, opened 2017-07-10
Fixed GetHullBounds

WPMGPRoSToTeMa:
Motivated by https://github.com/ValveSoftware/halflife/pull/1717.
This also fixes a bug with roof boost glitch (idk how to name it properly, I found it [here](https://forums.alliedmods.net/showthread.php?t=81877)). Video with demonstration https://youtu.be/9zkamv8shqM?t=7m55s from [here](https://forums.alliedmods.net/showthread.php?t=287955).
There is also [info](https://gist.github.com/WPMGPRoSToTeMa/2e6e0454654f9e4ca22ee3e987051b57) about `HUD_GetHullBounds` patching for the client. (this fixes twitching when you staying above crouching player https://youtu.be/0BUMoGDitVU)

### DIFF
--- a/regamedll/dlls/client.cpp
+++ b/regamedll/dlls/client.cpp
@@ -4601,9 +4601,46 @@ int EXT_FUNC ConnectionlessPacket(const struct netadr_s *net_from, const char *a
 	return 0;
 }
 
+// TODO: int -> qboolean or enum
 int EXT_FUNC GetHullBounds(int hullnumber, float *mins, float *maxs)
 {
-	return hullnumber < 3;
+#ifdef REGAMEDLL_FIXES
+	switch (hullnumber)
+	{
+	case 0: // Normal player
+		Q_memcpy(mins, (float*)VEC_HULL_MIN, sizeof(vec3_t));
+		Q_memcpy(maxs, (float*)VEC_HULL_MAX, sizeof(vec3_t));
+		return TRUE;
+	case 1: // Crouched player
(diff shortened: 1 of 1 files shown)


regamedll/dlls/client.cpp
```diff
@@ -4601,9 +4601,46 @@ int EXT_FUNC ConnectionlessPacket(const struct netadr_s *net_from, const char *a
...
+	case 0: // Normal player
+		Q_memcpy(mins, (float*)VEC_HULL_MIN, sizeof(vec3_t));
+		Q_memcpy(maxs, (float*)VEC_HULL_MAX, sizeof(vec3_t));
+		return TRUE;
+	case 1: // Crouched player
+		Q_memcpy(mins, (float*)VEC_DUCK_HULL_MIN, sizeof(vec3_t));
+		Q_memcpy(maxs, (float*)VEC_DUCK_HULL_MAX, sizeof(vec3_t));
+		return TRUE;
+	case 2: // Point based hull
+		Q_memcpy(mins, (float*)Vector(0, 0, 0), sizeof(vec3_t));
```
theAsmodai: maybe memset?

WPMGPRoSToTeMa: We can refactore and optimize it later. It is also called only once, so there is no difference.

leantipatico:
Can u help me with adding a command for putting on/off bug boost, in which you can go through walls with jump in duck, because we play HNS Boost and we need this thing in that mode?

PR #173, opened 2017-07-15
Fixed bug with lastinv (fix from 7561)

WPMGPRoSToTeMa:
Originally reported [here](https://github.com/ValveSoftware/halflife/issues/1815).
This bug is not reproducable in CS by default. To reproduce it you can enable knife drop, then use this command sequence `drop weapon_knife;drop;lastinv`.

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -6536,7 +6536,10 @@ BOOL EXT_FUNC CBasePlayer::__API_HOOK(RemovePlayerItem)(CBasePlayerItem *pItem)
 		pev->viewmodel = 0;
 		pev->weaponmodel = 0;
 	}
-	else if (m_pLastItem == pItem)
+#ifndef REGAMEDLL_FIXES
+	else
+#endif
+	if (m_pLastItem == pItem)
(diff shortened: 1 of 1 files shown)


regamedll/dlls/player.cpp
```diff
@@ -6536,7 +6536,7 @@ BOOL EXT_FUNC CBasePlayer::__API_HOOK(RemovePlayerItem)(CBasePlayerItem *pItem)
 		pev->viewmodel = 0;
 		pev->weaponmodel = 0;
 	}
-	else if (m_pLastItem == pItem)
+	if (m_pLastItem == pItem)
```
WPMGPRoSToTeMa: Forgot about `REGAMEDLL_FIXES`.

Vaqtincha:
test plugin
`
#include <amxmodx>
#include <hamsandwich>

public plugin_init() {
	RegisterHam(Ham_CS_Item_CanDrop, "weapon_knife", "CBasePlayerItem_CanDrop")
}

public CBasePlayerItem_CanDrop(const pItem)
{
    SetHamReturnInteger(true)        
    return HAM_SUPERCEDE
}`

PR #179, opened 2017-08-18
Fix some of C4 related bugs.

In-line:
In a while I will add new bugfixes, but it's fine to merge now. And please don't squash commits I want to get better stats :trollface: 

### Credits

- https://www.youtube.com/watch?v=RV-utOqHg_k - bug info
- https://forums.alliedmods.net/showthread.php?t=171831 - bug info
- https://t.me/devcs - some suggestions and bug reports
- @Vaqtincha - for suggestions
- Valve - stable buggy code

### DIFF
--- a/regamedll/dlls/ggrenade.cpp
+++ b/regamedll/dlls/ggrenade.cpp
@@ -1232,10 +1232,15 @@ void CGrenade::C4Think()
 
 		g_pGameRules->m_bBombDropped = FALSE;
 
+#ifndef REGAMEDLL_FIXES
 		if (pev->waterlevel != 0)
-			UTIL_Remove(this);
+			UTIL_Remove(this); // Causes infinite round
 		else
+#endif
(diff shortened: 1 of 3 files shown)


Phantomsq:
So, will these changes make strelba more comfortable? I hope, they will🙏

regamedll/dlls/ggrenade.cpp
```diff
@@ -1272,6 +1281,10 @@ void CGrenade::C4Think()
 					TheBots->OnEvent(EVENT_BOMB_DEFUSE_ABORTED);
 				}
 			}
+#ifdef REGAMEDLL_FIXES
+            else
+                SET_CLIENT_MAXSPEED(pPlayer->edict(), 1);
```
In-line: Very questionable for custom plugins that depend on old behavior, but this should be fine.

PR #180, opened 2017-08-19
Add cvar for radio icon

In-line:
https://youtu.be/qx1A4HDzhPA
I <3 Valve.
Thanks to @wopox1337 for idea


WPMGPRoSToTeMa:
There is a solution without removing this sprite.

In-line:
@WPMGPRoSToTeMa I will be happy if you create PR for that, but in any case somebody maybe wants to remove that icon: )

AndyMcNab:
As I stated in https://github.com/ValveSoftware/halflife/issues/1814 ; the mp_show_voice_icons variable and logic is already in CS and other multiplayer games after HalfLife.  I would just like the few lines needed to allow this to be merged back into the HalfLife 1 code. Thanks

PR #184, opened 2017-09-24
Add mp_max_intermission_time cvar

In-line:
Basically this gives opportunity to change intermission behavior for server administrators. Imagine situation when 32 noobs are not pressing any key and shouting that map change is going too long.

Credits (for idea)
- @wopox1337
- @Fant1kUA


WPMGPRoSToTeMa:
Why do we need a new cvar? Administrators can't control `mp_chattime`?

In-line:
When I faced this problem, I thought like this:
1. Intermission before map change is common behavior for a long time.
2. There is gonna be some addon developer, that relied on this behavior.
3. Same as 2, but for players that love intermission.

So it is better to leave a decision to the server administrator. I will add additional check to block SVC_INTERMISSION if mp_max_intermission_time is zero.

WPMGPRoSToTeMa:
@In-line may be better add `mp_intermission_auto_end` or something like this?

In-line:
@WPMGPRoSToTeMa I don't think that it's necessary, but if somebody else support that idea I see no reason for resistance :)
I think it will complicate things, if somebody wants `SVC_INTERMISSION` with 0 timeout, he or she can just set `mp_max_intermission_time` to `0.0001`

Issue #185, opened 2017-09-27
Tracking issue


In-line:
ValveSoftware/halflife#278

simeonniko:
Is this already fixed? https://forums.alliedmods.net/showpost.php?p=1463034&postcount=15

Daniele386:
No. 18 seems to be [dreamstalker/rehlds#558](https://github.com/dreamstalker/rehlds/issues/558), maybe other can confirm this too

Daniele386:
Maybe some Fixes from https://forums.alliedmods.net/showthread.php?t=171831 can be implemented?

Daniele386:
I can confirm/reproduce No. 23 

I can confirm/reproduce **No. 13 only on original gamedll**; it seems to be fixed on ReHLDS/ReGameDLL (I had AMXX disabled, so it's not AMXX related; I used dual berretas to reload as it takes very long)

fred0r:
think i didnt see it on the list - would be nice if flashes which explode behind players doesnt (under certain circumstances) fullflash them.
this amxmodx already seems to fix this:
https://forums.alliedmods.net/showpost.php?p=930803&postcount=42

PR #188, opened 2017-09-29
Fix dead defuser progress bar issues

In-line:
I wasn't able to reproduce bugs  5 and 6 in #185, but I guess it's possible, so better to cancel progress bar when defuser dies.


simeonniko:
https://github.com/ValveSoftware/halflife/issues/94

PR #200, opened 2017-11-05
Make C4 more responsible

In-line:
This will allow to faster cancel progress bar on the defuser. As well as other stuff in `C4Think` will became more precise (like bomb exploding, detection of the defuse/plant in air and so on).

### DIFF
--- a/regamedll/dlls/ggrenade.cpp
+++ b/regamedll/dlls/ggrenade.cpp
@@ -1,5 +1,11 @@
 #include "precompiled.h"
 
+#ifdef REGAMEDLL_FIXES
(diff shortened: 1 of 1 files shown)


regamedll/dlls/ggrenade.cpp
```diff
@@ -1,5 +1,11 @@
 #include "precompiled.h"
 
+#ifdef REGAMEDLL_FIXES
+	#define NEXT_DEFUSE_TIME 0.033f
```
WPMGPRoSToTeMa: `const` or `constexpr`?

PR #204, opened 2017-11-08
Fix #203

### LINKED ISSUE #203: Fix Knife Secondary Attack Animation
I'm requesting fix knife secondary attack animation. This bug is also in original mp.dll. When calling secondary attack of knife, it doesnt setting m_flTimeWeaponIdle, so if u used primary attack, after a while using secondary attack will disable its animation and will start idle anim.

In-line:


In-line:
@theAsmodai :shipit:

theAsmodai:
@In-line
squirrel

PR #235, opened 2017-12-12
Adjust bomb angles for righthanded users (resolves #218)

### LINKED ISSUE #218: Bad bomb angles (#207)
Related to https://github.com/s1lentq/ReGameDLL_CS/pull/207

Since most people use right hand ingame, bomb feel reversed about the angle. The right is at the left and so on.

That could be easly fix by changing the angle value, but i have no time to test this.

In-line:
Since `cl_righthand 1` is the most popular option we must adjust ourselves for it (look for the wires after the bomb is planted).


WPMGPRoSToTeMa:
We can detect if `cl_righthand` is `0` or `1` for specific player.

PR #258, opened 2018-01-30
Fixed: Flashbang should remain the active item after a throw

KaidoRen:


Freeman-AM:
Isn't that a big change in flashbang gameplay ?
I mean good players are used to the current flash system, aka switch to get back the 2nd flash.
That option should be at cvar discretion.

PR #260, opened 2018-02-02
Update version requirement for VS and ICC

In-line:
https://github.com/s1lentq/ReGameDLL_CS/blob/9374fced908d64d8dbc90c878605925b151078fa/regamedll/build.gradle#L135

https://software.intel.com/en-us/articles/c14-features-supported-by-intel-c-compiler
https://msdn.microsoft.com/en-us/library/hh567368.aspx


IgnacioFDM:
We should update VS project files and why not platform toolset

In-line:
Can somebody who is using VS test this?

PR #266, opened 2018-02-11
Spawn protection API and cvar (mp_respawn_immunitytime).

wopox1337:
# Added a new CVar: "mp_respawn_immunitytime"
**Description**: Specifies the players defense time after respawn. (in seconds).
**Values**:
> "0" - disabled;
> ">0.00001" - time delay to remove protection

**Visualization**:
mp_respawn_immunitytime


regamedll/dlls/API/CSPlayer.cpp
```diff
@@ -477,3 +477,25 @@ EXT_FUNC void CCSPlayer::StartDeathCam()
 {
 	BasePlayer()->StartDeathCam();
 }
+
+EXT_FUNC void CCSPlayer::SetProtectionOnSpawn()
```
In-line: You shouldn't implement this in the API. `CSPlayer` should be treated like proxy between gamedll and user code.

regamedll/dlls/API/CSPlayer.cpp
```diff
@@ -477,3 +477,25 @@ EXT_FUNC void CCSPlayer::StartDeathCam()
...
+{
+	CBasePlayer *pPlayer = BasePlayer();
+
+	pPlayer->pev->takedamage = DAMAGE_NO;
+	pPlayer->pev->rendermode = kRenderTransAdd;
+	pPlayer->pev->renderamt = 100.0;
```
In-line: It is better to create hookchain and left choice effect to the user.

regamedll/dlls/player.cpp
```diff
@@ -4375,6 +4375,15 @@ void EXT_FUNC CBasePlayer::__API_HOOK(PreThink)()
 		m_flDisplayHistory |= DHF_ROUND_STARTED;
 	}
 
+#ifdef REGAMEDLL_ADD
+	if (spawnprotectiontime.value > 0
+		&& CSPlayer()->m_bProtected
```
In-line: `m_bProtected` is redundant. Just check `m_flRemoveProtectionPending`

regamedll/dlls/API/CSPlayer.cpp
```diff
@@ -477,3 +477,25 @@ EXT_FUNC void CCSPlayer::StartDeathCam()
...
+}
+
+EXT_FUNC void CCSPlayer::RemoveProtectionOnSpawn()
+{
+	CBasePlayer *pPlayer = BasePlayer();
+
+	pPlayer->pev->takedamage = DAMAGE_AIM;
+	pPlayer->pev->rendermode = kRenderNormal;
+
+	m_bProtected = false;
```
In-line: Set `m_flRemoveProtectionPending` to `0`. `m_bProtected` is redundant

regamedll/public/regamedll/API/CSPlayer.h
```diff
@@ -79,13 +79,17 @@ class CCSPlayer: public CCSMonster {
...
+	virtual void SetProtectionOnSpawn();
+	virtual void RemoveProtectionOnSpawn();
 
 	CBasePlayer *BasePlayer() const;
 
 public:
 	char m_szModel[32];
 	bool m_bForceShowMenu;
 	float m_flRespawnPending;
+	float m_flRemoveProtectionPending;
```
In-line: Something like `m_flSpawnProtectionStart` seems to be better

regamedll/dlls/API/CSPlayer.cpp
wopox1337: later, as a plug-in ReAPI, using API from ReGameDLL that still need to be implemented

regamedll/dlls/player.cpp
wopox1337: m_bProtected - displays the status in the client protection in the ReAPI call, and is more convenient to use. Otherwise, you will have to enter the concept of spawnprotectiontime.value = -1, which would extend the concept of state, which is not a convenient solution.

justgo97:
add a cvar for render option ?

wopox1337:
> add a cvar for render option ?

@justgo97 What exactly do you want to change in render with CVar? More specifically, please.

LevShisterov:
Not sure why you renamed that cvar, mp_spawnprotectiontime was a better option, IMHO.

wopox1337:
@LevShisterov I would not like to confuse users who are trying to find in Google descriptions of the functionality. As well as following the names of GameDLL CS:GO.
https://developer.valvesoftware.com/wiki/List_of_CS:GO_Cvars

regamedll/dlls/player.h
```diff
@@ -343,6 +343,8 @@ class CBasePlayer: public CBaseMonster {
 	virtual BOOL AddPlayerItem(CBasePlayerItem *pItem);
 	virtual BOOL RemovePlayerItem(CBasePlayerItem *pItem);
 	virtual int GiveAmmo(int iAmount, const char *szName, int iMax = -1);
+	virtual void SetProtectionOnSpawn();
+	virtual void RemoveProtectionOnSpawn();
```
In-line: Protection isn't removed on spawn, so maybe `RemoveProtection` or `RemoveSpawnProtection`?

regamedll/dlls/player.h
```diff
@@ -343,6 +343,8 @@ class CBasePlayer: public CBaseMonster {
 	virtual BOOL AddPlayerItem(CBasePlayerItem *pItem);
 	virtual BOOL RemovePlayerItem(CBasePlayerItem *pItem);
 	virtual int GiveAmmo(int iAmount, const char *szName, int iMax = -1);
+	virtual void SetProtectionOnSpawn();
```
In-line: If you think about creating new API it is better to introduce argument, like `void SetProtectionOnSpawn(float fProtectionTime)` which is self explanatory. I guess at that case `SetProtectionOnSpawn(-1)` can mean infinite time. And I guess renaming to something like `SetSpawnProtection` should be better.

regamedll/dlls/player.cpp
```diff
@@ -9516,3 +9525,20 @@ bool CBasePlayer::__API_HOOK(CanSwitchTeam)(TeamName teamToSwap)
 
 	return true;
 }
+
+EXT_FUNC void CBasePlayer::SetProtectionOnSpawn()
```
In-line: This shouldn't be `EXT_FUNC`

LevShisterov:
I don't think that different action (no damage vs kick) would make a big difference here. If this is still confusing you can use mp_spawn_protection_time. At least "respawn" isn't right, because protection is present even at first spawn. And people already got accustomed to word "protection", IMHO.

wopox1337:
> At least "respawn" isn't right, because protection is present even at first spawn.

Hmm, right. I would not like to confuse the user with a variety of CVar names.
For CS:GO cvar called mp_spawnprotectiontime - performs still another action.

In-line:
cssdk and reapi should receive updates too

justgo97:
> @justgo97 What exactly do you want to change in render with additional CVar? More specifically, please.

cvar for render option on/off

and also : 

cvar for render type
cvar for render color rgb

wopox1337:
> @wopox1337
cvar for render option on/off

and also :

cvar for render type
cvar for render color rgb

@justgo97 I was planning to add in ReGameDLL basic functionality for the protection of the player on the spawn. 
And the render (player) in the base case should be a bit transparent. The API is made to be able to diversify the default functionality. This rendering system is used in CS: GO, I think it would be nice to add to our engine. 

For example, using the API will be possible to implement the removal of protection from the movement of the player. As well as setting the render color of the command.

lespaul64:
Actually the render effect is not needed at all, you just ruin 3rd-party render mod. Make spawn protection is enough. As far as i know HLDM don't have render effect too.

In-line:
@lespaul64 I agree with you too, but people don't.

PR #272, opened 2018-02-24
Remove rendering from spawn protection.

In-line:
Related to #266

> Actually the render effect is not needed at all, you just ruin 3rd-party rendering mod. Make spawn protection is enough. As far as i know HLDM don't have render effect too.

I agree with @lespaul64, because render effect is now uncustomizable and mod-breaking. It feels ideologically bad.

I think it's better to receive some community feedback until doing the final decision
@wopox1337 @WPMGPRoSToTeMa @s1lentq


justgo97:
mayby add a cvar to turn render OFF/ON ?

PR #279, opened 2018-03-30
Fix bug, mp_fraglimit is less than zero

wopox1337:
Why?
BUG: When user set mp_fraglimit < 0 - server will reboot every map when server have players.


In-line:
What if fraglimit is 0?

wopox1337:
@In-line All right. When CVar set to < 0 - we called a bug

PR #294, opened 2018-05-12
GCC port and new defines for new ARM port

a1batross:
A working port to GCC 7.3. 

I also have added two new defines:
* HAVE_SSE enabled automatically when compiler is enabled SSE support. It will be used for updated ARM port of ReGameDLL_CS.
* NO_AMXX_HACKS disables ASM inline. It will be used for updated ARM port of ReGameDLL_CS.

### DIFF
--- a/regamedll/dlls/saverestore.cpp
+++ b/regamedll/dlls/saverestore.cpp
@@ -28,6 +28,8 @@
 
 #include "precompiled.h"
 
+static constexpr int CSaveRestoreBuffer::m_Sizes[];
+
 TYPEDESCRIPTION g_EntvarsDescription[] =
 {
 	DEFINE_ENTITY_FIELD(classname, FIELD_STRING),
--- a/regamedll/dlls/saverestore.cpp
+++ b/regamedll/dlls/saverestore.cpp
@@ -28,7 +28,10 @@
 
 #include "precompiled.h"
 
+// this should be defined(otherwise link will fail), but MSVC doesn't require it
+#ifndef _MSC_VER
 constexpr int CSaveRestoreBuffer::m_Sizes[];
+#endif
 
 TYPEDESCRIPTION g_EntvarsDescription[] =
 {
--- a/regamedll/dlls/saverestore.cpp
+++ b/regamedll/dlls/saverestore.cpp
@@ -28,7 +28,7 @@
 
 #include "precompiled.h"
 
-static constexpr int CSaveRestoreBuffer::m_Sizes[];
+constexpr int CSaveRestoreBuffer::m_Sizes[];
 
 TYPEDESCRIPTION g_EntvarsDescription[] =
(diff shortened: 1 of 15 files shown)


a1batross:
As I am afraid to do any changes in Gradle, because I don't know it, I've published a CMakeLists.txt used to compile this to Gist here:
https://gist.github.com/a1batross/3b4e1ecc34c7a29646b6acc05a8d90f5

regamedll/dlls/saverestore.cpp
```diff
@@ -28,6 +28,11 @@
 
 #include "precompiled.h"
 
+// this should be defined(otherwise link will fail), but MSVC doesn't require it
+#ifndef _MSC_VER
+constexpr int CSaveRestoreBuffer::m_Sizes[];
```
WPMGPRoSToTeMa: I'm not sure, but looks like expression `m_Sizes[x]` (where `x` isn't constexpr) requires address evaluation and then this definition is required.

regamedll/dlls/util.h
```diff
@@ -323,7 +323,8 @@ inline T *UTIL_FindEntityInSphere(T *pStartEntity, const Vector &vecCenter, floa
 	pentEntity = FIND_ENTITY_IN_SPHERE(pentEntity, vecCenter, flRadius);
 	if (!FNullEnt(pentEntity))
 	{
-		return (T *)CBaseEntity::Instance(pentEntity);
+		// a1ba: can't use CBaseEntity::Instance here, because circular dependency in cbase.h
```
WPMGPRoSToTeMa: Can you give more info about this?

WPMGPRoSToTeMa:
> A working port to GCC 7.3.

Does it work on GCC 8.1?

regamedll/dlls/saverestore.cpp
WPMGPRoSToTeMa: This was fixed in C++17.

a1batross:
Yes, it will should work for any version of GCC supporting C++14. 

But I did not tried it. Unfortunately, GCC 8.1 is not yet arrived even in Ubuntu 18.04. :)

regamedll/dlls/util.h
a1batross: In cbase.h an util.h is included before CBaseEntity definition. So after preprocessing it's look like:
```
a template function with CBaseEntity::Instance call
...
...
...
actual CBaseEntity::Instance definition
```

In fact, Instance() method performs a nullptr check and returns private data of specified edict or world edict. 

This function already does a null check, so just GET_PRIVATE is enough.

regamedll/dlls/saverestore.cpp
a1batross: Ah, I see. Thanks for pointing me. :)

PR #297, opened 2018-06-15
(Add cvar) Friendly fire damage

spark512:
Intended to control the hardcoded damage done to teammates. Suggestions, improvements?


simosalis:
Adding "that support" (damage * 0.35 on teamates) for the HE grenades (and maybe C4) at the same occasion will be good too.
Because I don't see why the damage done by bullets is * 0.35 while not by HE...
Either nothing is supported, or either everything.

simosalis:
Adding "that support" (damage to teammates * 0.35) to the HE grenades (even C4) could be a nice addition/fix at the same occasion.
Because I don't see why the damage done by bullets is * 0.35 while not by HE.
Either all the weapons have that feature, or either none of it (logic).

eat1k:
@spark512 what about finishing this PR?

PR #301, opened 2018-07-16
Add new impulse 255, which gives you every weapon, just for debugging

a1batross:
Why 255? 0xFF is just a beautyful number.

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -4521,7 +4521,7 @@ void CBasePlayer::UpdateGeigerCounter()
 	m_flgeigerDelay = gpGlobals->time + 0.25;
 
 	// send range to radition source to client
-	range = byte(m_flgeigerRange / 4.0);//* 0.25);		// TODO: ACHECK!
+	range = byte(m_flgeigerRange / 4.0);// * 0.25);		// TODO: ACHECK!
 
 	if (range != m_igeigerRangePrev)
 	{
@@ -6028,9 +6028,9 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
 			ALERT(at_console, "Crediting %s with $16000\n", STRING(pev->netname));
 			break;
 #ifdef REGAMEDLL_ADD
-		case 255: 
+		case 255:
 		{
-			auto GiveFilledWeapon = [&](char* pszWeaponName) {
+			auto giveFilledWeapon = [&](const char* pszWeaponName) {
 				GiveNamedItem(pszWeaponName);
 				const WeaponInfoStruct *pInfo = GetWeaponInfo(pszWeaponName);
 				if (pInfo) {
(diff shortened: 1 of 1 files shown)


regamedll/dlls/player.cpp
```diff
@@ -6008,6 +6008,16 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
...
 
+#ifdef REGAMEDLL_ADD
+	auto GiveFilledWeapon = [&](char* pszWeaponName) {
+		GiveNamedItem(pszWeaponName);
+		const WeaponInfoStruct *pInfo = GetWeaponInfo(pszWeaponName);
+		if (pInfo) {
+			GiveAmmo(pInfo->maxRounds, pInfo->ammoName2);
+		}
+	};
+#endif // REGAMEDLL_ADD
```
WPMGPRoSToTeMa: Maybe create a block inside `case 255:` and move this lambda into it?
```C++
case 255: {
	auto GiveFilledWeapon = [&](char* pszWeaponName) {
		GiveNamedItem(pszWeaponName);
		const WeaponInfoStruct *pInfo = GetWeaponInfo(pszWeaponName);
		if (pInfo) {
			GiveAmmo(pInfo->maxRounds, pInfo->ammoName2);
		}
	};

	GiveFilledWeapon( "weapon_ak47" );
	GiveFilledWeapon( "weapon_aug" );
	// ...
}
```

a1batross: Yeah, indeed. That should be better.

regamedll/dlls/player.cpp
```diff
@@ -6027,6 +6027,47 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
 			AddAccount(16000);
 			ALERT(at_console, "Crediting %s with $16000\n", STRING(pev->netname));
 			break;
+#ifdef REGAMEDLL_ADD
+		case 255: 
+		{
+			auto GiveFilledWeapon = [&](char* pszWeaponName) {
```
WPMGPRoSToTeMa:
```suggestion
			const auto GiveFilledWeapon = [&](const char *pszWeaponName) {
```
and maybe change `GiveFilledWeapon` to `giveFilledWeapon`? (since it is a local variable)

regamedll/dlls/player.cpp
```diff
@@ -6027,6 +6027,47 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
...
+				GiveNamedItem(pszWeaponName);
+				const WeaponInfoStruct *pInfo = GetWeaponInfo(pszWeaponName);
+				if (pInfo) {
+					GiveAmmo(pInfo->maxRounds, pInfo->ammoName2);
+				}
+			};
+
+			GiveFilledWeapon( "weapon_ak47" );
```
WPMGPRoSToTeMa: Remove spaces around parentheses?

regamedll/dlls/player.cpp
```diff
@@ -4521,7 +4521,7 @@ void CBasePlayer::UpdateGeigerCounter()
 	m_flgeigerDelay = gpGlobals->time + 0.25;
 
 	// send range to radition source to client
-	range = byte(m_flgeigerRange / 4.0);//* 0.25);		// TODO: ACHECK!
+	range = byte(m_flgeigerRange / 4.0);// * 0.25);		// TODO: ACHECK!
```
WPMGPRoSToTeMa: Unrelated change.

regamedll/dlls/player.cpp
```diff
@@ -6027,6 +6027,48 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
 			AddAccount(16000);
 			ALERT(at_console, "Crediting %s with $16000\n", STRING(pev->netname));
 			break;
+#ifdef REGAMEDLL_ADD
+		case 255:
+		{
+			auto giveFilledWeapon = [&](const char* pszWeaponName) {
```
WPMGPRoSToTeMa: `const auto` doesn't fit here?

WPMGPRoSToTeMa: `const char *pszWeaponName` please.

WPMGPRoSToTeMa: Maybe rename `pszWeaponName` to `weaponName`?

regamedll/dlls/player.cpp
```diff
@@ -6027,6 +6027,48 @@ void CBasePlayer::CheatImpulseCommands(int iImpulse)
...
+				GiveNamedItem(pszWeaponName);
+				const WeaponInfoStruct *pInfo = GetWeaponInfo(pszWeaponName);
```
WPMGPRoSToTeMa: Consider `const WeaponInfoStruct &info`, so `info.maxRounds` and `info.ammoName2`.

regamedll/dlls/player.cpp
a1batross: It fixes code highlighting in GNU Nano. 

I did not mentioned it, as it just a comment.

regamedll/dlls/player.cpp
a1batross: Don't think so. `auto` main feature is to make code shorter, in this case `const` it doesn't make sense, as it's just a lambda.

regamedll/dlls/player.cpp
a1batross: GetWeaponInfo return value intended to be a pointer, isn't it?

regamedll/dlls/player.cpp
WPMGPRoSToTeMa: @a1batross and even without `const` this variable can't be reassigned since lambdas with capture don't have assignment operator (but assignment operator is available for stateless lambdas in C++20).

regamedll/dlls/player.cpp
WPMGPRoSToTeMa: @a1batross can you move this change to a new PR?

regamedll/dlls/player.cpp
WPMGPRoSToTeMa: Yes, but you can "convert" it to a reference by dereferencing it.

a1batross: And still, check for null pointer? %)

WPMGPRoSToTeMa: @a1batross can it return `nullptr` for this code?

PR #306, opened 2018-08-30
Do not ignore thirdparty bots in humans_join_team cvar checks.

jeefo:
Game dll considers third party bots as humans when dealing with humans_join_team cvar, thus making those bots to join team specified in this cvar.


s1lentq:
this the checks move to CBasePlayer::IsBot so that yapb are determined as bot for all code and 3rd-party stuff's.
I guess it should be better way.

PR #328, opened 2018-11-29
Added new CVar: "mp_block_cmd_drop" - to disable drop cmd from player.

wopox1337:
It need for so many mods and DeathMatch gameplay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Archive's bin directory contains 2 subdirectories, 'bugfixed' and 'pure'
 | mp_item_staytime              | 300     | -   | -            | Time to remove item that have been dropped from the players. |
 | mp_legacy_bombtarget_touch    | 1       | 0   | 1            | Legacy func_bomb_target touch. New one is more strict. <br/>`0` New behavior<br/>`1` Legacy behavior|
 | mp_respawn_immunitytime       | 0       | 0   | -            | Specifies the players defense time after respawn. (in seconds).<br/>`0` disabled<br/>`>0.00001` time delay to remove protection |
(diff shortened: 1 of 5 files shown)


wopox1337:
tested. Well works now.
image
image

wopox1337:
@WPMGPRoSToTeMa [i added this CVar earlier mp_item_staytime](https://github.com/s1lentq/ReGameDLL_CS/commit/411761c8d885fd521adfb5ac506618ef8846258e). 

Now i added cvar - to block CMD (Command from player).

WPMGPRoSToTeMa:
@wopox1337 `mp_weaponstay` cvar is boolean in HL.

WPMGPRoSToTeMa:
@wopox1337 maybe just `mp_block_drop` or `mp_block_weapon_drop`/`mp_block_item_drop`?

wopox1337:
@WPMGPRoSToTeMa The item can be dropped not only by player cmd but also by death or replacement with other weapons. What makes concrete the meaning of the name

WPMGPRoSToTeMa:
@wopox1337 then maybe something like `mp_block_drop_by_player` or `mp_block_drop_by_player_self`?

simosalis:
Maybe at the same occasion, extend/replace that feature by adding a CVar like "mp_block_cmds", working via flags (abcd...), where each flag allow to block an internal command.
For example, "a" for "say", "b" for "say_team", etc.
There is a limit with that (abcdefghijklmnopqrstuvwxyz), 26 commands, or 32 with the use of extras flags not related to letters ("|" for example) but we can probably regroup some commands to use the same flag (radio1, radio2... linked to one, etc.).

Such addition will give a better control over the default commands available to clients and provided by the game (as blocking the "vote" command, etc.).
Just a suggestion... then it's not long to implement.

PR #333, opened 2019-01-04
Remove hardcoded client speed set when wpn_c4 arming.

wopox1337:


regamedll/dlls/wpn_shared/wpn_c4.cpp
```diff
@@ -366,5 +372,5 @@ void CC4::Use(CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TYPE useType, f
 
 float CC4::GetMaxSpeed()
 {
```
s1lentq:
```
#ifdef REGAMEDLL_FIXES
	if (m_bStartedArming)
		return 1.0f;
#endif
```

regamedll/dlls/wpn_shared/wpn_c4.cpp
```diff
@@ -156,7 +160,9 @@ void CC4::PrimaryAttack()
 
 			if (m_fArmedTime <= gpGlobals->time)
 			{
+#ifndef REGAMEDLL_FIXES
 				if (m_bStartedArming)
```
WPMGPRoSToTeMa: Can it be just removed?

fant1kua: @WPMGPRoSToTeMa yes. Because before it we also checked this member. Please take a look on it https://github.com/s1lentq/ReGameDLL_CS/blob/8b0ec7a627c204b59e3c6db5202ebc4fda5c60da/regamedll/dlls/wpn_shared/wpn_c4.cpp#L121. It else case of if statement where we check m_bStartedArming in negative state. That's why it will be always true. And there is nothing that can change it.

s1lentq: @wopox1337 pls can you remove this block-if too

PR #339, opened 2019-02-01
Fixes

Vaqtincha:


eat1k:
Is it possible to do mp_buy_anywhere work on not default maps? I've tested mp_buy_anywhere 1. We still need to create buyzone icon so the player will be able to use "buy" cmd.

PR #340, opened 2019-02-14
Add CanHolster check

CrazY770:
Hello, just a little addition so player won't switch weapon if active item can't be holstered.


s1lentq:
Closed by 2defb8a

PR #347, opened 2019-03-10
Added cvar to fix pushable velocity on use

voed:
For some reason player was pushing along with func_pushable on pressing use.
Example here: https://www.youtube.com/watch?v=idAsaXfBRfI


regamedll/msvc/Новая папка/ReGameDLL.vcxproj
```diff
@@ -1,1672 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
WPMGPRoSToTeMa: LOL.

WPMGPRoSToTeMa:
How does this fix behave with slow pushables?

voed:
> How does this fix behave with slow pushables?
No idea. Where i can find them to test?

regamedll/msvc/Новая папка/ReGameDLL.vcxproj
voed: ¯\\_(ツ)_/¯

WPMGPRoSToTeMa:
@voed hmmmmm, I can't find anything slow...

It is also bugged in HL and behaves just like in CS when use slowdown is disabled. With use slowdown enabled player move speed isn't crazy, but pushables speed is still crazy.

Can you investigate it deeper? You can use `boot_camp` map as an example, it is runnable in CS (spawn as terrorist), you maybe will need to break several crates to move one of them. I think the goal would be to make these crates slow when pulling them.

voed:
@WPMGPRoSToTeMa you are right, tested on pure and bugfixed HL, bug doesnt reproduce only because of use slowdown, and reproduced on boot_camp map in CS

Vaqtincha:
voed dolbayob :D

RauliTop:
Taking advantage about this...

It will be a great idea fix func_pushable position every new round:
https://forums.alliedmods.net/showthread.php?p=616291

Mistrick:
> Taking advantage about this...
> 
> It will be a great idea fix func_pushable position every new round:
> https://forums.alliedmods.net/showthread.php?p=616291
Already fixed.
https://github.com/s1lentq/ReGameDLL_CS/blob/a732fad31bd4dea4e265cb700843eec683eafbee/regamedll/dlls/func_break.cpp#L910-L930

WPMGPRoSToTeMa:
I found [something interesting](https://github.com/FWGS/XashXT/blob/2700894a877b6b34bef8c69177a4904d90f318b6/server/func_break.cpp#L961-L969) in XashXT. Can you test it?

voed:
> I found [something interesting](https://github.com/FWGS/XashXT/blob/2700894a877b6b34bef8c69177a4904d90f318b6/server/func_break.cpp#L961-L969) in XashXT. Can you test it?

Well, it works completely different and buggy, and does not even solve the problem.
I can explain better in telegram if you want

Issue #345, opened 2019-03-04
ReGameDLL 5.7.0.333 remove fix of boost roof bug


FLYskills360:
Already completed !
Working as bugfixed version, but you can use cvar for BUG BOOST
mp_bug_boost_disable "1" = NOT HAVE A BOOST BUG
mp_bug_boost_disable "0" = BOOST BUG IS AVAILABLE
Special for HideNSeek Mode
Link : http://teamcity.rehlds.org/viewType.html?buildTypeId=ReGameDLL_Publish&guest=1&branchSelecto
Edited by stfkolev

RauliTop:
What's the reason of removing a bugfix? 🤔

FLYskills360:
stfkolev How i can put this ? I see u understand me and i am so grateful can you give me ready file i can't understand these codes

FLYskills360:
Where is mp.dll ?

stfkolev:
> Where is mp.dll ?

Learn to use the git correctly.http://teamcity.rehlds.org/viewLog.html?buildId=16365&buildTypeId=ReGameDLL_Publish&guest=1

FLYskills360:
Already completed !
Working as bugfixed version, but you can use cvar for BUG BOOST
mp_bug_boost_disable "1" = NOT HAVE A BOOST BUG
mp_bug_boost_disable "0" = BOOST BUG IS AVAILABLE
Special for HideNSeek Mode

popilas:
Hello, im use regamedll-dist-5.20.0.492-dev version and boost dont work for like hns servers

PR #348, opened 2019-03-14
Added mp_bug_boost_disabled cvar

stfkolev:
Pull requested as part in the stead of https://github.com/s1lentq/ReGameDLL_CS/issues/345

**Reason to merge with master:**
It is widely used technique in every hide'n'seek mode boost server.


regamedll/dlls/gamerules.h
```diff
@@ -872,6 +872,19 @@ inline bool HasRoundInfinite(int flags = 0)
 	return false;
 }
 
+inline bool isBugBoostDisabled(int flags = 0) 
+{
+	#ifdef REGAMEDLL_ADD
+		if(bug_boost_disabled.string[0] == '1')
```
Mistrick: why this type of check?

stfkolev: Why not? Can it be written better?

Mistrick: check by value member
what if i set cvar value "01"?

stfkolev: @Mistrick I don't think it will have any mistakes, because it gets the first character only, but okay. So you say if I do it like that: `bug_boost_disabled.value == 1` it'll be better?

stfkolev:
As it was tested, it seemed to do not care of the value from the cvar. It allows boosting whatsoever. Currently investigating the issue.

FLYskills360:
> Pull requested as part in the stead of #345
> 
> **Reason to merge with master:**
> It is widely used technique in every hide'n'seek mode boost server.

Can you sent me ready file please :)

FLYskills360:
Where is mp.dll file ?

FLYskills360:
I am SO GRATEFUL for this, thank you very much !!!

s1lentq:
Closes by 8c85aee

PR #351, opened 2019-04-07
Buffer overflow fix

2010kohtep:
A small fix for some UTIL functions that helps prevent shellcode injection using BSP files.

### DIFF
--- a/regamedll/dlls/util.cpp
+++ b/regamedll/dlls/util.cpp
@@ -1248,7 +1248,8 @@ void UTIL_StringToVector(float *pVector, const char *pString)
 	char tempString[128];
 	int j;
 
-	Q_strcpy(tempString, pString);
+	Q_strncpy(tempString, pString, sizeof(tempString));
+	tempString[sizeof(tempString) - 1] = '\0';
 
 	pstr = tempString;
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
Is this from the latest update? If no then it should be wrapped with `REGAMEDLL_FIXES`. Moreover you can remove string copying since the string is not modified.

2010kohtep:
Yes, this fix is also included in the Valve update, which was released a few days ago. I preferred to repeat the original code taken from **client.so,** that's why ```REGAMEDLL_FIXES``` is not needed too.

regamedll/dlls/util.cpp
```diff
@@ -1248,7 +1248,8 @@ void UTIL_StringToVector(float *pVector, const char *pString)
 	char tempString[128];
 	int j;
 
-	Q_strcpy(tempString, pString);
+	Q_strncpy(tempString, pString, sizeof(tempString));
+	tempString[sizeof(tempString) - 1] = '\0';
```
s1lentq: @2010kohtep please use `Q_strlcpy` instead of this code

PR #352, opened 2019-04-10
Fix player's velocity on pushable use

voed:
More info in #347, i've broken that pr @_@


WPMGPRoSToTeMa:
Cvar???

voed:
@s1lentq decided that cvar is unnecessary

Karaulov:
@voed On all maps it used as is, with bug. Only new maps can use this fix. Possible need add cvar to enable this bug again?

voed:
> On all maps it used as is, with bug

No, its not.

Karaulov:
> No, its not.

When this bug has been added?

PR #368, opened 2019-05-14
Add damage source for explode

Karaulov:
Screenshot (destroy func_breakable with explode) , kill hostage, kill enemy, suicide. :)
Y7lKn0Q 1


Karaulov:
For test, works

PR #369, opened 2019-05-22
Add CVar: mp_afk_bomb_drop_time.

wopox1337:
Player that have never moved sience they last move will drop the bomb after this amount of time. (in seconds).


Karaulov:
Player drop c4 even if not afk. (Just stay on place and aiming)

wopox1337:
@Karaulov when player Aiming:
```
(m_pPlayer->pev->button & IN_ATTACK2))
{
	SecondaryAttack();
}
```

Karaulov:
Every 5 second? 

And if player moving with SHIFT/CTRL, c4 dropped.

Karaulov:
On next round, player drop bomb))))) at round start)))

Karaulov:
@wopox1337 On next round, player drop bomb))))) at round start)))

PR #381, opened 2019-06-09
 [WIP] New CVar mp_buy_anywhere

wopox1337:
When set, players can buy anywhere, not only in buyzones. 0 = default. 1 = both, 2 = Terrorists, 3 = CT.


dist/game.cfg
```diff
@@ -262,3 +262,12 @@ ff_damage_reduction_other "0.35"
...
 mp_afk_bomb_drop_time "0"
 
+// When set, players can buy anywhere, not only in buyzones.
+// 0 - default
+// 1 - both teams
+// 2 - only Terrorists team
+// 3 - only CT team
+//
+// Default value: "0"
+// mp_buy_anywhere "0"
```
Karaulov: Commented :( Sadness :(

Karaulov: @wopox1337 does not match the remaining CVARs formatting style :)

wopox1337: Not understand. details pls

Karaulov: Sorry, bad google translator! 

Пишу что не соответствует форматированию в этом файле :) _( Другие CVAR не закомментированы :) )_ ну вижу что исправлено уже https://github.com/s1lentq/ReGameDLL_CS/pull/381/commits/622193ccb9b08d3da0916d958e596e8e82096339

StevenKal:
It will be better to change the correspondance of the CVar's settings.
In other terms:
0 - Default (feature disabled).
1 - Terrorists only.
2 - Counter-Terrorists only.
3 - Both teams.

The logic behind most of the CVars should be like:
0 - Disabled.
1 - A bit enabled.
2 - A bit more enabled.
3 - A bit more more enabled (fully...).

Or, maybe a short rework, where -1 is the default game behavior and 0 disable buying.

WPMGPRoSToTeMa:
What is the reason for specific team restriction?

wopox1337:
@WPMGPRoSToTeMa such values are in CS:GO and maybe someone will need customization

StevenKal:
@WPMGPRoSToTeMa:
The main reason is "customization", more you can do, more it's welcomed, especially with some people who could have different wish/requests (can be frequent). So at least they could set the settings like they want.
This can be used/useful according to the layout/shape of a specific map (gameplay wanted), where the Terros can not be able to buy, but find armouries around them, and where CTs can not find armouries in their sector, but by able to buy weapons/items.
As example, on a map with a castle or a museum exposing weapons (like in the "Demolition Man" movie with Stallone/Snipes), where terros have invested the place!
By the way, this is a setting allowable by default, with the "info_map_parameters" entity (m_iBuyingStatus) and with the CGameRules (m_b<T|CT>CantBuy).

RauliTop:
> Or, maybe a short rework, where -1 is the default game behavior and 0 disable buying.

I don't think it's a good idea. If you want to disable buy just change mp_buytime.
However, I think your new values to the cvar are better, of course.

> What is the reason for specific team restriction?

Customization using mods.
Without mods, yes, there shouldn't any reason.

PR #394, opened 2019-07-29
Radio enchance cvars

wopox1337:
Added a new CVar's for Radio messages control.
```
 "mp_radio_timeout" - Delay between player Radio messages;
 "mp_radio_maxinround" - Maximum Radio messages count for player per round.
```


regamedll/dlls/gamerules.h
```diff
@@ -886,6 +888,15 @@ inline float CGameRules::GetItemKillDelay()
 #endif
 }
 
+inline float CGameRules::GetRadioTimeout()
+{
+#ifdef REGAMEDLL_ADD
+	return radio_timeout.value;
```
wopox1337: I guess there's no need a clamp() or min().
Usage example: 
1) https://github.com/s1lentq/ReGameDLL_CS/blob/e9a732f813605efeecf43edd7780f9837069e779/regamedll/dlls/client.cpp#L2096
2) https://github.com/s1lentq/ReGameDLL_CS/blob/e9a732f813605efeecf43edd7780f9837069e779/regamedll/dlls/client.cpp#L2089

regamedll/dlls/player.cpp
```diff
@@ -5202,7 +5202,11 @@ void EXT_FUNC CBasePlayer::__API_HOOK(Spawn)()
 	m_flLastTalk = 0;
 	m_flIdleCheckTime = 0;
 	m_flRadioTime = 0;
+#ifdef REGAMEDLL_ADD
+	m_iRadioMessages = radio_maxinround.value;
```
etojuice: I suggest doing explicit type conversion, like this:
`m_iRadioMessages = int(radio_maxinround.value);`

PR #404, opened 2019-08-18
Fix: func_healthcharger, func_recharge, func_break, func_tank

Vaqtincha:


regamedll/dlls/mapinfo.cpp
```diff
@@ -74,7 +74,7 @@ void CMapInfo::KeyValue(KeyValueData *pkvd)
 	}
 	else if (FStrEq(pkvd->szKeyName, "bombradius"))
 	{
-		m_flBombRadius = (float)Q_atoi(pkvd->szValue);
+		m_flBombRadius = Q_atof(pkvd->szValue);
```
s1lentq: Behavior changed, float rounded here, so you have 2 way
1. Leave this
2. Wrap in a macro REGAMEDLL_FIXES

regamedll/dlls/h_battery.cpp
```diff
@@ -67,11 +67,6 @@ void CRecharge::Restart()
 	UTIL_SetSize(pev, pev->mins, pev->maxs);
 	SET_MODEL(ENT(pev), STRING(pev->model));
 
-	int armorValue = (int)gSkillData.suitchargerCapacity;
-	if (pev->armorvalue != 0.0f) {
-		armorValue = (int)pev->armorvalue;
-	}
-
```
s1lentq: Why was this removed?

regamedll/dlls/maprules.cpp
```diff
@@ -185,7 +185,7 @@ void CGameText::Spawn()
 {
 #ifdef REGAMEDLL_FIXES
 	// Don't allow entity triggering itself
-	if (FStrEq(pev->target, pev->targetname))
```
s1lentq: This change is pointless, because `pev->target` or `pev->targetname` are `QString` class that those have operator `operator const char *()`

regamedll/dlls/h_battery.cpp
Vaqtincha: Not used. 

Not necessary. (Func "Recharge")

regamedll/dlls/maprules.cpp
Vaqtincha: Ok.

Vaqtincha:
:cry:

PR #410, opened 2019-09-06
add prevent jump flag to iuser3 data

fl0werD:


RauliTop:
It should be REGAMEDLL_ADD not REGAMEDLL_FIXES

PR #415, opened 2019-09-14
Add "give"  client cheat command (e.g give weapon_ak47) like in csgo

Vaqtincha:


Vaqtincha:
https://github.com/s1lentq/ReGameDLL_CS/pull/301

PR #416, opened 2019-09-14
Add CVar mp_show_scenarioicon

Vaqtincha:
Condition-Zero style scenario icon near round timer.

Безымянный


RauliTop:
**If the purpose is to add this new message to 1.6 ignore this...**

According to the original code (don't know why only sends message when playing CZero).

It should be (after refactoring):
```
if (!AreRunningCZero() || show_scenarioicon.value == 0.0f)
     return;
```

First time was wrong.
After refactoring was wrong again.


Good check (before refactoring):
```
if (AreRunningCZero() && show_scenarioicon.value != 0.0f)
{
    //...
}
```

s1lentq:
> If the purpose is to add this new message to 1.6 ignore this...

Why not? CS 1.6 can handles this message.

RauliTop:
> CS 1.6 can also handle this message.

I didn't know that.

Now all clear, nothing against this request.

PR #428, opened 2019-10-03
New CVar mp_give_player_c4

wopox1337:
```
// Whether this map should spawn a C4 bomb for a player or not.
// 0 - disabled
//
// Default value: "1"
mp_give_player_c4 1
```

P.S: [from CS:GO](https://totalcsgo.com/command/mpgiveplayerc4)


StevenKal:
So it's for playing the bomb map but without taking in account the objectives which has to be performed, right?
I think if we go on this path a CVar with flags (like "mp_round_infinite" is) will be welcomed, in order to have more control on objectives. So we could choose or not on assasinations maps if we want the map grants a VIP, spawn hostages, also remove/restore static decals related to those objectives (A/B/(+) decals for bomb target, hostage rescue zone, etc.).

regamedll/dlls/player.cpp
```diff
@@ -9734,7 +9734,11 @@ bool EXT_FUNC CBasePlayer::__API_HOOK(GetIntoGame)()
 		Spawn();
 		CSGameRules()->CheckWinConditions();
 
-		if (!CSGameRules()->m_flRestartRoundTime && CSGameRules()->m_bMapHasBombTarget && !CSGameRules()->IsThereABomber() && !CSGameRules()->IsThereABomb())
+		if (!CSGameRules()->m_flRestartRoundTime && CSGameRules()->m_bMapHasBombTarget && !CSGameRules()->IsThereABomber() && !CSGameRules()->IsThereABomb()
+#ifdef REGAMEDLL_ADD
+			&& give_player_c4.value
```
justgo97: Maybe just make one check inside the **GiveC4** function?

wopox1337: No. Its bad for thrid-party plugins

zoruamonster:
@s1lentq possible to add new features ?

mp_give_player_c4 2 = give bomb for player who successful plant bomb in prev round and/or give bomb only for best players (by score) for example for top5 players

wopox1337:
@zoruamonster Better do this by plugins

zoruamonster:
> @zoruamonster Better do this by plugins

yes??

wopox1337:
Da

PR #429, opened 2019-10-03
New CVar mp_weapons_allow_map_placed

wopox1337:
```
// When set, map weapons (located on the floor) will be shown.
// NOTE: Effect will work after round restart.
// 0 - hide all map weapons
//
// Default value: "1"
mp_weapons_allow_map_placed 1
```
P.S: from [CS:GO](https://totalcsgo.com/command/mpweaponsallowmapplaced)


RauliTop:
@s1lentq @wopox1337 this feature has the values inversed.

From CSGO:
0 - do not remove weapons from map
1 - clear all weapons when round starts

PR #435, opened 2019-10-08
Implement RG_CBasePlayer_DropIdlePlayer hook

d3m37r4:
It will expand the capabilities of the standard afk checker to the needs of the user, without need to block standard functionality of ReGameDLL.

[duplicate](https://github.com/s1lentq/ReGameDLL_CS/pull/409)


Vaqtincha:
все сломал.

Issue #3, opened 2015-11-29
Too fast firing with dual elites


WPMGPRoSToTeMa:
With this code https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/dlls/wpn_shared/wpn_elite.cpp#L126 dual elites can fire about 13 times per second. It's too fast, isn't it?
You can test it with this:
```
alias f "+attack;wait;-attack;wait"
alias f2 "f;f"
alias f4 "f2;f2"
alias f8 "f4;f4"
alias f16 "f8;f8"
alias f32 "f16;f16"
alias f64 "f32;f32"
bind mouse1 f64
```

s1lentq:
thanks! good to know.
i think enough

```
flCycleTime -= 0.078;
```

WPMGPRoSToTeMa:
Yeah, this value is best balanced.

satnatantas:
Oh I see, just need "pure" version.

satnatantas:
I'd like to make a CVAR for that, will you accept it if I program it? I think that it should be the same as in pure library because firing speed is already balanced with mad recoil (which is twice the recoil displayed on screen, bullets actually go even higher). You are just killing fun because many servers now use bugfixed ReGame.

voed:
@s1lentq can we revert this? It breaks game balance and seems completely illogic.
For example, glock has 0.05 cycletime, usp has 0.075, so why two guns cannot shoot faster than one?🤔

satnatantas:
The "fix" has also a terrible side effect being that the client will display shots at requested speed without using bullets unless the client has very small network latency,

WPMGPRoSToTeMa:
https://coub.com/view/1mlp7k

Vaqtincha:
fix strelba ot voedlol

satnatantas:
I appreciate this reversion!

satnatantas:
This weapon is almost never used, it is next to impossible to control it at higher firing rate, and the bullet marks are not even correct with cl_lw 0 on any server and it has parallax unlike any other weapon - these are all the issues you need to deal with when you use it. Please do not ruin the only fun weapon who nobody uses.

ValveSoftware/halflife#2822

satnatantas:
pussies

PR #473, opened 2019-12-20
Revert wpn_elite cycletime fix

voed:
Explained here https://github.com/s1lentq/ReGameDLL_CS/issues/3#issuecomment-558133236


satnatantas:
This weapon is almost never used, it is next to impossible to control it at higher firing rate, and the bullet marks are not even correct with cl_lw 0 on any server and it has parallax unlike any other weapon. Please do not ruin the only fun weapon who nobody uses.

https://github.com/ValveSoftware/halflife/issues/2822

wopox1337 (commented):
nope

wopox1337:
a

wopox1337:
I think at the moment ReGameDLL has a normal shooting speed, at least based on CS:GO (which is supported by Valve).

Video: https://youtu.be/w5fJrNsYa-U
[name](https://youtu.be/w5fJrNsYa-U)

PR #479, opened 2019-12-22
Grenadefixes

Vaqtincha:


fl0werD:
В чом прекол

etojuice:
> 3c6bacb
> 51c86ad

IMO these features should be made as optional

Vaqtincha:
No.

3c6bacb - forget code
51c86ad - bug

Vaqtincha:
grenade can't deployed on pickup (if the player did not have a weapon)

https://www.youtube.com/watch?v=uSxIaeqMlMU

dystopm:
> grenade can't deployed on pickup (if the player did not have a weapon)
> 
> https://www.youtube.com/watch?v=uSxIaeqMlMU

Hi @Vaqtincha , got one question, that hegrenade you wanted to pickup, was originally an armoury_entity or weaponbox? Because it matters, a lot
In case answer is weaponbox, you first had ONLY a hegrenade in your inventory, and then you dropped it and try to pick it up?

In this cases of bugfixes approaches a little of context is always needed. The better communication the best, we can't guess by just looking at the code because it can involve big chains of structures and flows that can be easily broken.

(so this way this will not turn into a real "circus")

Vaqtincha:
@dystopm bug only with armory_entity. on armory touch the player receives via givenameditem and CanDeploy is called before the player is filled with ammo (m_pPlayer->m_rgAmmo[m_iPrimaryAmmoType])

Vaqtincha:
And this is imperceptible when the player has any other weapon.

dystopm:
I'm not understanding then. GiveNamedItem creates, Spawn assigns m_iDefaultAmmo, touching called on DefaultTouch where AddPlayerItem is called, then CBasePlayerWeapon::AddToPlayer which gives ammo on ExtractAmmo, gun is linked to player and then inside FShouldSwitchWeapon there is the first call to CanDeploy which returns true because ammo was given on ExtractAmmo/AddPrimaryAmmo. Am I forgetting something? 🤔

Vaqtincha:
this is a problem with weaponbox and not armory_entity (many years have passed and I forgot) but the problem is the same (CanDeploy is called before the player gets ammo)

dystopm:
> this is a problem with weaponbox and not armory_entity (many years have passed and I forgot) but the problem is the same (CanDeploy is called before the player gets ammo)

Well between these 2 entity there's a huge difference.

armoury_entity gives you ammo on AddWeapon (which is ExtractAmmo), it doesn't need to get into AddDuplicate chain. so this was discarded from before. Was going to ask you if you were using an edited fork or something.

weaponbox gives you gun first, ammo then, but that can be fixed in two ways:
- Instead of giving weaponbox ammo from hegrenade (PackAmmo on CreateWeaponBox), ammo should go to m_iDefaultAmmo - only on EXHAUSTIBLE weapons
- Checking THIS particular scenario when player has no guns -> I prefer this one, because the first one can alter some modded behaviour

Stay tuned

Vaqtincha:
Ahahah. "A huge difference between weaponbox and armory_entity." Thanks, I didn't know. 🙃 There is no need to be smart here. I know very well how they work.

dystopm:
>  There is no need to be smart here. I know very well how they work.

> @dystopm bug only with armory_entity. on armory touch the player receives via givenameditem and CanDeploy is called before the player is filled with ammo (m_pPlayer->m_rgAmmo[m_iPrimaryAmmoType])

¿?

I couldn't say you know "well" how this works if you were ignoring the behavior of both entities and your proposed fix is only changing CanDeploy to "true" randomlessly. There IS a need to be smart here, what there is no need for is to be spicy.

Issue #514, opened 2020-02-21
[bug] cvars mp_forcecamera, mp_forcechasecam doesn't work


wopox1337:
Always available only spec for temmates

regener:
+1
Same since last update.

PR #507, opened 2020-02-06
Fix fadetoblack issues

wopox1337:
close #501 
close #506

PR #518, opened 2020-03-13
Fix force camera issue

fant1kua:
Fix https://github.com/s1lentq/ReGameDLL_CS/issues/514


RauliTop:
@s1lentq

s1lentq:
@RauliTop #507 was canceled so this PR makes no sense

PR #532, opened 2020-04-23
Player fixes

Vaqtincha:


regamedll/dlls/player.cpp
```diff
@@ -7806,8 +7806,10 @@ CBaseEntity *EXT_FUNC CBasePlayer::__API_HOOK(DropPlayerItem)(const char *pszIte
 
 #ifdef REGAMEDLL_FIXES
 		// No more weapon
-		if ((pev->weapons & ~(1 << WEAPON_SUIT)) == 0) {
+		if ((pev->weapons & ~(1 << WEAPON_SUIT)) == 0) 
+		{
 			m_iHideHUD |= HIDEHUD_WEAPONS;
+			m_iFOV = DEFAULT_FOV;
```
justgo97, line 7812: This should be added to RemoveItem in case player get his awp stripped while zooming

PR #570, opened 2020-10-03
Fix preventing +USEing buttons through walls

theshadowz:
In some cases, for example, on cs_mansion map flFraction doesn't return 1.0f value (Sometime returns 0.84...f or 0.91...f). Thats why i decided to use 0.8f value

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -3988,7 +3988,18 @@ bool CanSeeUseable(CBasePlayer *me, CBaseEntity *pEntity)
 			}
 		}
 	}
+#ifdef REGAMEDLL_FIXES
+	else if (FClassnameIs(pEntity->pev, "func_button"))
+	{
+		Vector vecOrigin;
+		vecOrigin = VecBModelOrigin(pEntity->pev);
 
+		TraceResult Tr;
+		UTIL_TraceLine(me->pev->origin, vecOrigin, ignore_monsters, ignore_glass, me->edict(), &Tr);
+
+		return (Tr.flFraction >= 0.8f);
+	}
+#endif
 	return true;
 }
 
@@ -4126,7 +4137,6 @@ void CBasePlayer::PlayerUse()
 	{
 		if (!useNewHostages || CanSeeUseable(this, pObject))
 		{
-			// TODO: traceline here to prevent +USEing buttons through walls
 			int caps = pObject->ObjectCaps();
 
 			if (m_afButtonPressed & IN_USE)


regamedll/dlls/player.cpp
```diff
@@ -3988,7 +3988,18 @@ bool CanSeeUseable(CBasePlayer *me, CBaseEntity *pEntity)
...
 		}
 	}
+#ifdef REGAMEDLL_FIXES
+	else if (FClassnameIs(pEntity->pev, "func_button"))
+	{
+		Vector vecOrigin;
+		vecOrigin = VecBModelOrigin(pEntity->pev);
+
+		TraceResult Tr;
+		UTIL_TraceLine(me->pev->origin, vecOrigin, ignore_monsters, ignore_glass, me->edict(), &Tr);
```
fl0werD, line 3998: Why not eye instead of me->pev->origin?

theshadowz: Do you mean view_ofs?

fl0werD: Yes, see this https://github.com/s1lentq/ReGameDLL_CS/blob/cc732c876b209d18922fddf7c87f9d4246ea7789/regamedll/dlls/player.cpp#L3969

theshadowz: My fault. Already done. Thanks

fl0werD: And check with 1.0 fraction now.

fl0werD: It seems to me that needs a cvar.

theshadowz: No, the problem i described above, unfortunatelly, still exists with 1.0f value. The value 0.8f fixes this problem now too

theshadowz: > It seems to me that feature needs a cvar.

Good idea

fl0werD: With REGAMEDLL_ADD macro.

theshadowz: Done. It looks really better for me now

RauliTop:
Great fix with a cvar to prevent problems in some maps functionality.

Waiting to the merge!

theshadowz:
@RauliTop, it's a little bit wrong realisation. Unfortunately, dont have a time to rewrite it

Nord1cWarr1or:
What about defuse bomb through wall?

theshadowz:
> What about defuse bomb through wall?

The fix is also concerns about defusing a bomb through wall. But as i said above, dont have a time to do right realisation because of long departure. You're all welcome to finish the PR

hajimura:
Closing the PR because of https://github.com/s1lentq/ReGameDLL_CS/pull/687

PR #584, opened 2020-11-25
Reset m_flNextFollowTime before trying to find next target after previous target disconnects

etojuice:


afwn90cj93201nixr2e1re:
.

PR #595, opened 2021-01-12
enhance CVar "mp_roundover"

Vaqtincha:
Maybe rename to mp_roundover_winner ?


Vaqtincha:
Not work!

PentalginPlus:
>// 1 - round end with Terrorists win

I understand the logic of assigning CVar values in this case, but I'm afraid that server owners can overlook the necessity of changing their map configs and will encounter issues after getting an update, because their maps or mods logic don't expect "Terrorists win"

PR #624, opened 2021-05-15
fix weapon_sg550

Vaqtincha:
bug https://youtu.be/PYq1BRg_fbo


Vaqtincha:
https://github.com/ValveSoftware/halflife/issues/2719

PR #633, opened 2021-05-25
fix `CBasePlayer::AddPointsToTeam`

francoromaniello:
index is also part of the team


wopox1337:
@francoromaniello Describe the purpose of these edits in more detail. At the moment it is not quite clear what it is for. 
Perhaps there are examples of maps and how they don't work correctly? Or are you creating a map with `game_score` entity yourself and not using it correctly?

Also, why these edits are not under `#ifdef REGAMEDLL_FIXES` ?

francoromaniello:
it's not a bug, it's my mistake how i interpret the name of the function.
Sorry for the late response.

PR #653, opened 2021-06-27
FIX: Unexpected behavior with `mp_forcerespawn`

FEDERICOMB96:
Basically there is an exploit (or bug?) where depending on mp_forcerespawn if his value is higher to 0 and lower than 1, when you respawn you have a brief window to change your team, if you succesfully manage to change your team between the mp_forcerespawn value and WITHOUT closing the change appearance menu, you won't die due to "m_fNextSuicideTime", that will trigger the change team but without actually changing your skin model, you will keep the enemy one but the team change will success.

Thanks @metita for helping me with this


wopox1337:
@FEDERICOMB96 @metita I confirm that there is a problem
It was possible to change the command without dying because of `m_fNextSuicideTime`.

As soon as the suicide was performed, it is spawned, then if you suddenly change the command, `ClientKill()` will not work, [because `m_fNextSuicideTime` will not allow.](https://github.com/s1lentq/ReGameDLL_CS/blob/fbbd6bd0838749753e10499a77477bf2fb2324b1/regamedll/dlls/client.cpp#L384-L386)

## Conclusion:
If [`ClientKill()` was called by GameDLL](https://github.com/s1lentq/ReGameDLL_CS/search?l=C%2B%2B&q=ClientKill) code (not by the game client, via "kill") - it should work without delay.
Engine (HLDS) calls `ClientKill()` (`pfnClientKill`) only in one case - [when the player client sends "kill"](https://github.com/dreamstalker/rehlds/blob/6a916d766b2f766ed9a45f9f008b3f7b737e3c89/rehlds/engine/host_cmd.cpp#L2697) and only then, here it needs the client's spam protection.

## Solution
Everywhere in the ReGameDLL where `ClientKill()` is called, it is necessary to set `m_fNextSuicideTime = 0.0f` before calling it.

Thank you for finding the problem.

wopox1337 (changes requested):
https://github.com/s1lentq/ReGameDLL_CS/pull/653#issuecomment-1354492105

wopox1337 (approved):

PR #663, opened 2021-08-16
Implement `CBasePlayer::Observer_SetMode` hook

Lopol2010:
related issue s1lentq/reapi#207

### DIFF
--- a/regamedll/dlls/API/CAPI_Impl.cpp
+++ b/regamedll/dlls/API/CAPI_Impl.cpp
@@ -186,6 +186,8 @@ GAMEHOOK_REGISTRY(CBaseEntity_FireBullets);
 GAMEHOOK_REGISTRY(CBaseEntity_FireBuckshots);
 GAMEHOOK_REGISTRY(CBaseEntity_FireBullets3);
 
+GAMEHOOK_REGISTRY(CBasePlayer_Observer_SetMode);
+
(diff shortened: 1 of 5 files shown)


StevenKal:
Value of "GetForceCamera" hook could possibly overwrite your custom mode, and if set to "CAMERA_MODE_SPEC_ONLY_FIRST_PERSON" (2) eye camera will be set internally.

Lopol2010:
Hi it seems in my server spectator restrictions work fine. Also i think if you have problems then you can set GetForceCamera return to any values?

StevenKal:
This can be pushed too (with adjustements regarding other hooks of fl0werD, or keep v5.21 and merge all [4 hooks] at once).

Lopol2010:
@StevenKal, Are you talk to me? If so, do you mean i need to resolve conflicts?

hajimura:
I think conflicts should be fixed anyway

StevenKal:
I was just refering to the "5.21" version upgrade, but I think while the maintainer/collaborators do not push a release in the "Releases" branch but keep adding/merging some new API functions, your hook could keep "5.21" as version and be added after fl0werD's hooks recently added (other API functions could even be added at same time and keep same version), no need to push to 5.22.

wopox1337 (approved):

PR #673, opened 2021-09-07
Disable thread-safe initialization for static local variables

jeefo:
Thread local storage (TLS) in shared libraries loaded by LoadLibrary() is not completely supported on Windows XP / Server 2003, causing game to crash if run on this platforms. But ReGamedDLL uses v140_xp platform toolset with GitHub Actions when building windows binary, assuming that XP/2003 must be supported.

This only affects VS2015+ builds, so we're not checking v120_xp toolset.

Probably the same should be applied for ReHLDS to support Windows XP properly.


s1lentq (approved):

StevenKal:
I love that!

PR #675, opened 2021-09-07
game.cfg: set CVar `sv_alltalk` to 1 by default

Nord1cWarr1or:
Why is this important?
Many users don't know about this CVar, and wonder why the players can't hear each other.
Just take a look at dev-cs or some other forums, people ask questions in revoice, VTC and gamedll threads "why the players can't hear each other?".
And a very, very small percentage of users use `sv_alltalk 0`.


Nord1cWarr1or:
A typo in the description of the second commit. sv_alltalk -> sv_allchat.

wopox1337:
The value in the `.cfg` must match the value in the description `README.md`, as well as the code. Rejected.

PR #686, opened 2021-10-08
New CVars: `sv_autobunnyhopping` and `sv_enablebunnyhopping`

aleeperezz16:
These commands work like in CS:GO: 
- `sv_autobunnyhopping` - allows the player to hold the jump button and jump automatically.
- `sv_enablebunnyhopping` - allows the player to bypass the max speed when jumping.

Maybe some API members can be added for 3rd party plugins and make auto-bunnyhopping per player.

### DIFF
--- a/dist/game.cfg
+++ b/dist/game.cfg
@@ -469,6 +469,7 @@ sv_allchat 0
 //
 // Default value: "0"
 sv_autobunnyhopping 0
+
 // Allow player speed to exceed maximum running speed
 // 0 - disabled (default behaviour)
 // 1 - enabled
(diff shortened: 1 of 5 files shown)


afwn90cj93201nixr2e1re:
.

afwn90cj93201nixr2e1re:
.

aleeperezz16:
> It’s all good, but why? It’s not flexible, sometimes you need to adjust velocity for specific events and/or players.

When I first came up with this, I had in mind to ease 3rd party plugins (like AMXX e. g.) as PlayerMove is not easy to control even with plugins. KreedZ mod helped in this idea as there are some maps called AXN that needs to bypass maxspeed limit when bhopping and also adds the autobhopping for these particular maps.

As you know, this is not necessary but a simple one line code feature for those who want, for a specific mod, autobhopping and don't want to waste their time making one modifying player's Pre/PostThink.
Why not making a simple feature easier for developers? I think that's why ReGameDLL is made for (also for bug fixing).

aleeperezz16:
> I like the idea, if we can add some members as you stated that would be pretty helpful tho.

If this is approved for a merge, I can make the API members.

PR #690, opened 2021-10-21
Modify `pev->frags` to `CBasePlayer::AddPoints` for ReAPI hooking

aleeperezz16:
So the idea is to change everytime `pev->frags` is used for adding o substracting frags to `CBasePlayer::AddPoints()` as this method is used for this.

On every `AddPoints()` call I manually set `bAllowNegativeScore` to `TRUE` as I don't find any purpose to don't let negative score to players. A CVar can make this a little more dynamic.


Vaqtincha:
it is better to add a new hook with one more argument "type" ex: AddFrags (player, count, type) enum fragType { Frag_kill, Frag_bomb, Frag_defuse, Frag_other, // ... }

aleeperezz16:
> it is better to add a new hook with one more argument "type" ex: AddFrags (player, count, type) enum fragType { Frag_kill, Frag_bomb, Frag_defuse, Frag_other, // ... }

Yeah, this is what I had in mind. But why not use same AddPoints? I don't think backward compatibility would break.

Issue #683, opened 2021-09-25
Add CVar `mp_plant_c4_anywhere`


JUNESYNGOTOFLEX:
implement this cvar from cs:go
this cvar like mp_buy_anywhere
8192.0

Vaqtincha:
Perebor.🙂

Maxi605:
mp_plant_anywhere sounds better

PR #694, opened 2021-10-25
Add CVar "mp_plant_c4_anywhere_delay" + API member.

StevenKal:
Related to #683.

Add CVar "mp_plant_c4_anywhere_delay" + API member CCSPlayer's "m_flPlantC4AnywhereDelay" to allow per-client setting.
I made work the CVar to support a delay (or none when set to "-1") from the round start, because this is a bit abused to have that feature active from beginning! Like teams spawns then 5 seconds later "The bomb has been planted!".
Then I have also added a message informing the player who wants to plant outside from a bomb site that "he has to wait * second(s)", but the problem with that is there is no localization string (#*) on the client (so no game translations), so this is discutable, despite I think this worth to inform him about the feature.
So maybe we can discuss about it if you devs are against or not, or, what the overall player base is thinking.

**About API member addition:**
The new member CCSPlayer's "m_flPlantC4AnywhereDelay" has "-2.0" as defaut "unconsidered value", this is in order to allow per-client support of "restriction" (if set to "0") instead of redirecting in such case to the CVar. The only workaround to support this kind of feature is to set the CVar to 0 & set per-client settings via the member, so it is easier the way I did.
Also, I have updated ReGameDLL_CS's API minor because any addition to API even in the class "CCSEntity" worth to increment this (to know the new member requires a specific API version of ReGameDLL_CS, despite the "commits number" could be used as workaround for such "tiny" thing).

**Extra notes:**
Also added "EXPORT" label to 2 functions of the class "CHalfLifeMultiplay" in order to allow easy retrieval of the function address via symbol under Windows (like under Linux).
I could also make them "virtual" but I prefered to do not in order to avoid breaking already existing virtual IDs listing ("HasRoundTimeExpired" + "IsBombPlanted" being at the suite...) & AMXX's ReAPI module using full CSSDK (and not only the "pure public ReAPI interfaces" like my own [AMX's ReAPI](https://www.amxmod.net/forum/showthread.php?tid=1522) module).

The kind of per-client API member could be added for a lot of other features (mp_buytime, mp_buy_anywhere, mp_forcerespawn, mp_freeforall, etc.), I basically have plans to do it because this allows customization per-client & so interesting, like if you want to enable FFA for only 1 client (like I have this on my public classic server, I can TK my teammates (the dumb ones!) this counts as frag! Hehehehehe!).

**PS**: I did not see someone already pushed similar PR [here](https://github.com/s1lentq/ReGameDLL_CS/pull/692) like 1 hour before, and I had mine ready from yesterday, but mine is more "advanced" and does not deal with signals (not sure if it is the best but at least client carrying the C4 can know if he is on a bomb site... as indication, and this is also that Maxi605 said so I guess no red blink icon under CS:GO right? I can not check I do not have the game). Feel free to choose which one you want! Or, extend the other one with the features of mine (as the API member).

### DIFF
--- a/regamedll/dlls/addons/item_airbox.cpp
+++ b/regamedll/dlls/addons/item_airbox.cpp
@@ -63,6 +63,7 @@ void CItemAirBox::Touch(CBaseEntity *pOther)
 
 		pev->nextthink = 0;
 		SetThink(nullptr);
+		pev->velocity = g_vecZero;
 	}
 }
 
@@ -70,6 +71,12 @@ void CItemAirBox::Restart()
 {
 	CArmoury::Restart();
 	UTIL_SetOrigin(pev, pev->oldorigin);
+	pev->velocity = g_vecZero;
+
+	if(m_flyup < 0)
+	{
+		m_flyup = -m_flyup;
+	}
 
 	if (m_flyup > 0 && m_delay > 0.01f)
 	{


wopox1337:
> With my code, you could have full control of the feature per each client, while with the "m_bCanPlantAnywhere" of the other PR you merged,, no.

I don't agree with you or I misunderstood you. Below is the code that you can control for each individual client.
```pawn
set_member(index, m_bPlantC4Anywhere, <boolean>)
```

> The function "CHalfLifeMultiplay::GetPlantC4AnywhereDelay" has been made for a reason, and its goal was to return "the final usable value related to the condition" (if client or not, if CCSPlayer's API value used or global one from CVar). …

At this point, I've taken a simpler route for this kind of functionality, which is unlikely to resonate widely in use.
I still question whether we need to make things so complicated in this simple (and probably useless) feature.

> But I can see we do not have the same "vision"…

And that's great! We can help each other learn different nuances of this difficult thing without getting into a conflict. :)

> The goal of this is to allow full customization per client (like per-client buy anywhere, C4 plant anywhere, etc.), so admins could easily customize all of this per client, really enjoyable.

You have a great idea. I suppose it would be worth discussing separately in another Issue. If you do that, you need to refactor many parts.

> I am a bit disappointed to submit something more enhanced…

The final decision cannot be made by me anyway. This is an open product, which is created by many people (so I wish). And I just don't have enough skilfulness to compete in some decisions (for their implementation). However, at this point, I make the simplest decision based on its relevance. In any case, we can always expand and fix the meager implementation as well as the API, thanks to cooperation. 

> …and see you prefered…

My subjective perception may differ greatly. Moreover, I do not claim to be "the smartest one here. I could be wrong :). Please worry less about such trivialities.

> my vision of having features "a little more than just basic" is not welcomed or shared, I will just not waste my time on this

I would be happy to bring more understanding people into the discussion of your endeavors. However, I don't guarantee that they will come and they are still interested in doing this. (But, to get their attention, it would be better to localize what we would ask them to ponder. (there are too many words - `TL;DR` trouble...)

For the next few times, let's try not to get bogged down in writing a memoir, but rather provide dry information to save time for anyone who is going to read it. A good tone would be to have a cold-blooded conversation of a substantive nature. For such long posts, it would also be good to format the right parts of the code with [Markdown](https://guides.github.com/features/mastering-markdown/). Moreover, for the accessibility of your text and thought to a wider audience, it would be better to write more concisely, using more common words. As far as I understand it, it would be easier to have conversations that way.

StevenKal:
> I don't agree with you or I misunderstood you. Below is the code that you can control for each individual client.
set_member(index, m_bPlantC4Anywhere, <boolean>)

I think you misunderstood me for this, or do not see the advantage of it!
As I said, a member like "m_iCanPlantC4Anywhere" can allow more control than a "m_bPlantC4Anywhere", if you set to "-1" to avoid considering the member (redirect/point to the CVar value for such client, just like if there were no API member, so in such case it's like setting/letting "m_bPlantC4Anywhere" to false), "0" to consider the member & point to "0" for such client (so disable the feature for the specific client while the CVar can be to 1 & enabled for the others, which, you can not do with a simple "m_bPlantC4Anywhere" boolean), and "1" to consider the member & point to "1" for such client.
This design of the API members is better than the standard/existing ones, it would be nice to update via this method, some others like "m_bAutoBunnyHopping", etc., but this will break the "offsetof", and so AMXX's ReAPI module using them, where it will need an to be recompiled with an updated CSSDK, besides, it already has its enums named the same as they are.
But I think for the future members we can do "my way" which is more customizable.

> The goal of this is to allow full customization per client (like per-client buy anywhere, C4 plant anywhere, etc.), so admins could easily customize all of this per client, really enjoyable.
You have a great idea. I suppose it would be worth discussing separately in another Issue. If you do that, you need to refactor many parts.

Thanks! But I will work on this later, this is just "something I have in mind in a TO DO list", because I have other things to do, but sometimes I like to change a bit & do other things (like submitting PRs for those good projects), the brain usually likes! Doing redundant things everytime can be exhausting & boring from a moment, surprise ourselves sometimes, with new tasks can be good!

For the rest, as I said on the "EDIT" part in the end, all is O.K. now (I could even remove some parts of the topic, but I hidden now as "resolved"), your idea of having a separate delay for the C4 is good, the problem is that I misunderstood you & I though you wanted to split "the C4 anywhere delay feature I made" into 2 CVars (which I considered stupid because an unique one is simpler & avoid "adding 1 more CVar..., & me sometimes I enrage myself a little bit quicker on this period, sorry), and since this was on the "anywhere topic", while the "mp_plant_c4_delay" was in fact, for a global feature being independant, which is, as I said on #698, a good idea.
Despite the name tell us this information properly, I think next time you can precise something like "this "mp_plant_c4_delay" must be independant & not only related to the plant anywhere feature", I would have understand it more.
But I also did not see the other messages on the other topic, as I usually only check when I have the small blue dot on the notifications, so a bit of my fault too! Haha! Bam! Bam! Bam! Spank on my ass!

About the C4 delay on the other topic, I quickly made a draft code yesterday at the evening, I will test it later, but I have added a new message for this, & also changed the "m_bPlantC4Anywhere" to work like I originaly wanted, since you did not pushed yet "an official release" we can change this, & please do not update AMXX's ReAPI module with this member till we did not pushed the C4 delay feature. Also do not push an official RG_CS release for now.

About my writting style, I will try to simplify texts, but you know this is stronger than me to write a lot in order to give details, hard to reduce my habit regarding this, besides since my main PC is on XP, the extras buttons on the top-right (quote, bold, etc.), do not work, I will have to list the "codes of them" in a .txt file in order to use them because I do not recall all, since this is not "bbcode style" but other codes.

PR #695, opened 2021-10-25
`RemoveSpawnProtection()` little fix

aleeperezz16:
So the actual problem of this happens when you hold any of the `IN_ACTIVE` buttons while you're dead and respawn afterwards. The icon from `StatusIcon` set doesn't get removed and even if you don't have protection anymore, the icon still stands on your HUD.
These commands can help replicate the issue:
`mp_respawn_immunitytime 3.0`
`mp_respawn_immunity_effects 1`
`mp_respawn_immunity_force_unset 1`
`mp_forcerespawn 1`

The "fix" I found is changing `MSG_ONE_UNRELIABLE` to `MSG_ONE` once it is set from `SetSpawnProtection()`.

I tried to replicate the same problem and never ever occured.


justgo97:
Have you tried instead changing this https://github.com/s1lentq/ReGameDLL_CS/blob/c30fe01f8e60ea1a6a4896bc9e0beb09ae0db580/regamedll/dlls/player.cpp#L10239 to **MSG_ONE_UNRELIABLE**  , maybe they work better when in the same channel?

aleeperezz16:
Yeah, I tested using that channel and sometimes got removed, others didn't or even not showed up the icon

wopox1337 (approved):

PR #707, opened 2021-11-12
Add cvar: mp_weapon_knockback

justgo97:
Allows to disable engine knockback behavior, useful for mods where they don't want this feature or want to implement their own knockback system like zombie mods...etc

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,7 +173,11 @@ jobs:
           fi
         shell: bash
 
-      - name: Build using Intel C++ Compiler 19.0
+      - name: Build using Intel C++ Compiler 19.0 (only for release)
+        if: |
+          github.event_name == 'release' &&
+          github.event.action == 'published' &&
+          startsWith(github.ref, 'refs/tags/')
         run: |
           rm -rf build-icc && CC=icc CXX=icpc cmake -B build-icc && cmake --build build-icc -j8
 
@@ -189,7 +193,7 @@ jobs:
       - name: Move files
         run: |
           mkdir -p publish/bin/linux32/cstrike/dlls
-          mv build-icc/regamedll/cs.so publish/bin/linux32/cstrike/dlls/cs.so
+          mv build-icc/regamedll/cs.so publish/bin/linux32/cstrike/dlls/cs.so 2>/dev/null || true
           mv build-gcc/regamedll/cs.so publish/cs-gcc.so
           mv regamedll/version/appversion.h publish/appversion.h
           mv dist/ publish/


justgo97:
@Vaqtincha why the downvote? 👀

hajimura:
Because recently some people makes shitty pull requests and this is one of them 🤡. tf is this? Who needs this? If you're interested in this idea, do it in your fork of the project or via amxx

justgo97:
This is a hardcoded thing in the regamedll, you can't reliably change it in amxx...

> Because recently some people makes shitty pull requests and this is one of them 🤡. tf is this? Who needs this? If you're interested in this idea, do it in your fork of the project or via amxx

hajimura:
TakeDamage_Pre hook -> get_entvar(id, var_velocity, m_iVelocity)
TakeDamage_Post hook -> set_entvar(id, var_velocity, m_iVelocity)

Vaqtincha:
@justgo97 Make it more functional and not just on / off. Or is it better to add api (player member) Ps: I would go even longer and make the slowdown (knockback) from the ammo type.

justgo97:
> TakeDamage_Pre hook -> get_entvar(id, var_velocity, m_iVelocity) TakeDamage_Post hook -> set_entvar(id, var_velocity, m_iVelocity)

That's not a clean method especially when you have plugins constantly checking/modifiying velocity, try to constantly print the velocity value and you will understand what I mean, not changing the velocity in the first place is cleaner.

justgo97:
> @justgo97 Make it more functional and not just on / off. Or is it better to add api (player member)
> 
> Ps: I would go even longer and make the slowdown (knockback) from the ammo type.

That's a good idea, But maybe in the future, for now, I just want a clean method to disable the engine from messing with the player velocity and the cvar is the only possible way to do it.

fl0werD:
It's useful for zombie modes, can you add api for 170?

justgo97:
> It's useful for zombie modes, can you add api for 170?

Well probably, but you can with this cvar disable the engine knockback and implement your own system in amxx especially if you have custom weapons and stuff.

Vaqtincha:
No, it doesn't work that way. You have to make it more versatile.
But it's still better to add a memeber.

justgo97:
> No, it doesn't work that way. You have to make it more versatile. But it's still better to add a member.

Well, That can be done in another PR, Also maintainers don't seem much active so complex PRs might be not a good idea at this time...

Vaqtincha:
@justgo97 Don't waste your precious time on this. They don't take your pr. Better develop your fork.

regamedll/dlls/game.cpp
```diff
@@ -163,6 +163,8 @@ cvar_t sv_autobunnyhopping               = { "sv_autobunnyhopping", "0", 0, 0.0f
 cvar_t sv_enablebunnyhopping             = { "sv_enablebunnyhopping", "0", 0, 0.0f, nullptr };
 cvar_t plant_c4_anywhere                 = { "mp_plant_c4_anywhere", "0", 0, 0.0f, nullptr };
 
+cvar_t weapon_knockback                 = { "mp_weapon_knockback", "1", 0, 0.0f, nullptr };
```
wopox1337, line 166:
```suggestion
cvar_t weapon_knockback                  = { "mp_weapon_knockback", "1", 0, 0.0f, nullptr };
```

regamedll/dlls/player.cpp
```diff
@@ -1143,8 +1143,13 @@ BOOL EXT_FUNC CBasePlayer::__API_HOOK(TakeDamage)(entvars_t *pevInflictor, entva
 		{
 			if (pev->velocity.Length() < 300)
 			{
-				Vector attack_velocity = (pev->origin - pAttack->pev->origin).Normalize() * 170;
-				pev->velocity = pev->velocity + attack_velocity;
+#ifdef REGAMEDLL_ADD
+				if (weapon_knockback.value != 0.0)
```
wopox1337, line 1147:
```suggestion
				if (weapon_knockback.value)
```

README.md
```diff
@@ -103,6 +103,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | sv_autobunnyhopping                | 0       | 0   | 1            | Players automatically re-jump while holding jump button.<br/>`0` disabled <br/>`1` enabled |
 | sv_enablebunnyhopping              | 0       | 0   | 1            | Allow player speed to exceed maximum running speed.<br/>`0` disabled <br/>`1` enabled |
 | mp_plant_c4_anywhere               | 0       | 0   | 1            | When set, players can plant anywhere, not only in bombsites.<br/>`0` disabled <br/>`1` enabled |
+| mp_weapon_knockback                | 0       | 0   | 1            | When set weapons will have the default knockback.<br/>`0` disabled <br/>`1` enabled |
```
wopox1337, line 106:
```suggestion
| mp_weapon_knockback                | 0       | 0   | 1            | When set, weapons will have the default knockback.<br/>`0` disabled <br/>`1` enabled |
```

dist/game.cfg
```diff
@@ -483,3 +483,10 @@ sv_enablebunnyhopping 0
 //
 // Default value: "0"
 mp_plant_c4_anywhere 0
+
+// When set weapons will have the default knockback.
```
wopox1337, line 487:
```suggestion
// When set, weapons will have the default knockback.
```

wopox1337 (changes requested):

justgo97:
> @justgo97 Don't waste your precious time on this. They don't take your pr. Better develop your fork.

Be patient 😄

Vaqtincha:
@justgo97 Don't get too excited. He is a rotten man. Trying to prove he exists here.

Vaqtincha:
Наверняка принимает этот бесполезный и не функциональный пр. Но это не оправдывает его неадекватность и высокомерность к отношению другим людям (ведь они потратили время). Отмазки "У миня нет вримени" да нет пару минут чтобы кликать энтер.

justgo97:
> @justgo97 Don't get too excited. He is a rotten man. Trying to prove he exists here.

Calm down, Your behavior is really toxic, This PR or others are not a high priority, It's not a critical bug fix or something.

Vaqtincha:
@justgo97 Even my critical bug fixes do not want to accept. Clown.

wopox1337:
1) It would be nice to adjust the multiplier (`0.0 .. 1.0`) rather than just the switch (`true .. false`).
2) This implementation is too simple, and does not leave the flexibility to handle a wider range of tasks. More flexible use should be considered.
3) Perhaps a better solution would be to put the entire code section into a separate function.
https://github.com/s1lentq/ReGameDLL_CS/blob/e613735900b4dca9a7fb094771653f50d18cdc9c/regamedll/dlls/player.cpp#L1144-L1155 
([or even more](https://github.com/s1lentq/ReGameDLL_CS/blob/e613735900b4dca9a7fb094771653f50d18cdc9c/regamedll/dlls/player.cpp#L1090-L1159))

justgo97:
I guess it's better to separate that into multiple PRs, Unfortunately, Currently I don't have the time to make more changes and test each case, I might only do minor and easy to test changes for now.

> 1. It would be nice to adjust the multiplier (`0.0 .. 1.0`) rather than just the switch (`true .. false`).
> 2. This implementation is too simple, and does not leave the flexibility to handle a wider range of tasks. More flexible use should be considered.
> 3. Perhaps a better solution would be to put the entire code section into a separate function.
>    https://github.com/s1lentq/ReGameDLL_CS/blob/e613735900b4dca9a7fb094771653f50d18cdc9c/regamedll/dlls/player.cpp#L1144-L1155
>    
>    ([or even more](https://github.com/s1lentq/ReGameDLL_CS/blob/e613735900b4dca9a7fb094771653f50d18cdc9c/regamedll/dlls/player.cpp#L1090-L1159))

RauliTop:
It will be good to have an option for disable this and apply our full-customized knockback system.

Also, I don't understand that 300 velocity magical number
and
m_flVelocityModifier = 0.65f;

why is inside there and not out that check?

RauliTop:
The implementation is too simple, but we really need this to disable CS default knockback system and apply our own.
Probably, instead of a cvar on/off, will be better a cvar to change that * 170 multiplier. Value 0 will be marked as off.

We can't have flexibility, having flexibility will make change large pieces of CS code, and it's not necessary because we can do it with plugins.
Why? knockback on CS is not always applied. It depends on:
- ShouldDoLargeFlinch (player ducking or not, some weapons/body part hit hasn't got knockback)
- Velocity < 300

So, please @wopox1337 reconsider this feature.

PR #714, opened 2022-01-05
README.md: Note about `impulse 255` command

UnkwUsr:
About this:
https://github.com/s1lentq/ReGameDLL_CS/blob/c002edd5b18a8408e299bc6cccfec2c7de56ba3d/regamedll/dlls/player.cpp#L6527-L6544


Vaqtincha:
Even such a harmless change will not be accepted "without tests" 🤣

Vaqtincha:
Типа "смотрите я оперативный" 😀

PR #716, opened 2022-01-08
remove buggy immunity status icon

Vaqtincha:


Vaqtincha:
Тестируйте хорошенько и не принимайте пр сразу. Лет где-то 3-4 надо тестировать.

StevenKal:
That shield icon for "protection" is not bad, and if the #695 solves the issue of it which keeps blinking after the end when it should not, just use it instead or remove it.

Vaqtincha:
@StevenKal No. I've only done bad things for cs.

StevenKal:
I am so sorry to disturb you in your "bad things work"! Hehehehehe!

Vaqtincha:
Pofig togda. 😁

PR #725, opened 2022-02-06
Add `max_id_range` CVar

Nord1cWarr1or:
This CVar allows you to define your own range to show player's nickname.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | sv_autobunnyhopping                | 0       | 0   | 1            | Players automatically re-jump while holding jump button.<br/>`0` disabled <br/>`1` enabled |
 | sv_enablebunnyhopping              | 0       | 0   | 1            | Allow player speed to exceed maximum running speed.<br/>`0` disabled <br/>`1` enabled |
 | mp_plant_c4_anywhere               | 0       | 0   | 1            | When set, players can plant anywhere, not only in bombsites.<br/>`0` disabled <br/>`1` enabled |
+| max_id_range                       | 2048    | 0   | 8192         | Defines the maximum visibility range of a player nickname.<br/>NOTE: Works if mp_playerid CVar value is not 2. |
 </details>
 
 ## How to install zBot for CS 1.6?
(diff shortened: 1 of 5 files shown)


wopox1337:
Need to handle this CVar with negative value too.

Nord1cWarr1or:
> I can't imagine a useful scenario for this change. It might be better to make it possible to control the distance through the API, but not the CVar.

Default value does not allow showing nickname on big maps, like de_abaddon.

s1lentq:
> > I can't imagine a useful scenario for this change. It might be better to make it possible to control the distance through the API, but not the CVar.
> 
> Default value does not allow showing nickname on big maps, like de_abaddon.

In this case, wouldn't it be better to use the `MaxRange` map value?

WPMGPRoSToTeMa:
> Default value does not allow showing nickname on big maps, like de_abaddon.

Could you please open an issue for that for better tracking? (a screenshot(s) or a video would be a good addition)

Nord1cWarr1or:
> > Default value does not allow showing nickname on big maps, like de_abaddon.
> 
> Could you please open an issue for that for better tracking? (a screenshot(s) and/or a video would be a good addition)

Sure

Nord1cWarr1or:
> In this case, wouldn't it be better to use the `MaxRange` map value?

Maybe so. But I think that some pople can watns to keep default behavoir?

StevenKal:
Might also relies on CVar "sv_zmax" instead of making a new one.
So in such case the maximum player visibility could be linked to the max "map visibility".
A new CVar allows customization, so it is always better, but I do not see much the point to customize that kind of stuff to a lower value (like 500 units).
And another name for the CVar with a common prefix might be preferable (like "mp_playerid_range_max", something like).

Nord1cWarr1or:
@s1lentq, @StevenKal, ok, I agree with you. I'll open another PR if I can.

StevenKal:
Was no need to close it, if you O.K. to use zmax, just update it with pointer of CVar "sv_zmax" (add a global "g_psv_zmax" in "game.<cpp|h>" files to get the pointer & replace "max_id_range.value" by "g_psv_zmax.value" or more safe by "(g_psv_zmax ? g_psv_zmax.value : 2048)")!
Then remove the other parts no needed in other files.

Issue #703, opened 2021-11-06
fast fire glitch


JUNESYNGOTOFLEX:
if you take sg552/aug and hold fire (with scope) you fire speed is slowly than usually, but if you flags ~FL_ONGROUND you fire speed faster than m249, fix please

aleeperezz16:
Is this a bug or it's intended? Conditions are made to take priority on firing when a player isn't on ground.

https://github.com/s1lentq/ReGameDLL_CS/blob/b979b5e84f36dc0eb870f97da670b700756217f1/regamedll/dlls/wpn_shared/wpn_aug.cpp#L81-L96

https://github.com/s1lentq/ReGameDLL_CS/blob/b979b5e84f36dc0eb870f97da670b700756217f1/regamedll/dlls/wpn_shared/wpn_sg552.cpp#L80-L95

RauliTop:
It seems to be a bug because they forget about check FOV at that weapons.

Fixed by this:
```
if (!(m_pPlayer->pev->flags & FL_ONGROUND)) 
{ 
 	if (m_pPlayer->pev->fov == DEFAULT_FOV) 
 	{ 
 	 	AUGFire(0.035 + (0.4 * m_flAccuracy), 0.0825, FALSE); 
 	} 
 	else 
 	{ 
 	 	AUGFire(0.035 + (0.4 * m_flAccuracy), 0.135, FALSE); 
  	}
} 
else if (m_pPlayer->pev->velocity.Length2D() > 140) 
{ 
 	if (m_pPlayer->pev->fov == DEFAULT_FOV) 
 	{ 
 	 	AUGFire(0.035 + (0.7 * m_flAccuracy), 0.0825, FALSE); 
 	} 
 	else 
 	{ 
 	 	AUGFire(0.035 + (0.7 * m_flAccuracy), 0.135, FALSE); 
  	}
} 
else if (m_pPlayer->pev->fov == DEFAULT_FOV) 
{ 
 	AUGFire(0.02 * m_flAccuracy, 0.0825, FALSE); 
} 
else 
{ 
 	AUGFire(0.02 * m_flAccuracy, 0.135, FALSE); 
 }
```

RauliTop:
What do you think @wopox1337?

Vaqtincha:
@RauliTop she's is dead

wopox1337:
[...]

PR #734, opened 2022-02-21
Fix: 'fast fire glitch' at AUG/SG552

RauliTop:
Bug is caused because of not checking weapon's FOV in some conditions.

close https://github.com/s1lentq/ReGameDLL_CS/issues/703

comment from @wopox1337 
> If we keep changing the shared code of weapons for the server side, at some point in time we will get more and more out of sync with the client side (prediction) and this will adversely affect the overall experience of the game for the average player.
> 
> I propose to leave this behavior "as is".

Well, if you don't want to merge it. Let be open to documentation.

### DIFF
--- a/regamedll/dlls/wpn_shared/wpn_aug.cpp
+++ b/regamedll/dlls/wpn_shared/wpn_aug.cpp
@@ -80,11 +80,25 @@ void CAUG::PrimaryAttack()
 {
 	if (!(m_pPlayer->pev->flags & FL_ONGROUND))
 	{
-		AUGFire(0.035 + (0.4 * m_flAccuracy), 0.0825, FALSE);
+		if (m_pPlayer->pev->fov == DEFAULT_FOV)
+		{
+			AUGFire(0.035 + (0.4 * m_flAccuracy), 0.0825, FALSE);
+		}
+		else
+		{
+			AUGFire(0.035 + (0.4 * m_flAccuracy), 0.135, FALSE);
+		}
 	}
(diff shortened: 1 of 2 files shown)


Vaqtincha:
KickBack should also be smaller when the player aimed (fov 55)

RauliTop:
> KickBack should also be smaller when the player aimed (fov 55)

Why? kickback isn't based on FOV

Vaqtincha:
Up

wopox1337 (approved):
As I said, there are problems with wpn_shared behavior prediction discrepancy, but they only affect clients whose pings are above 70. In general, it's not critical, I think it's possible to make this fix in ReGameDLL.

Video: https://youtu.be/-SLAXhhGwu0
[name](https://youtu.be/-SLAXhhGwu0)

wopox1337:
btw [CS:GO code](https://github.com/perilouswithadollarsign/cstrike15_src/blob/f82112a2388b841d72cb62ca48ab1846dfcc11c8/game/shared/cstrike15/weapon_sg552.cpp#L62-L76):
```cpp

void CWeaponSG552::PrimaryAttack()
{
	CCSPlayer *pPlayer = GetPlayerOwner();
	if ( !pPlayer )
		return;

	bool bZoomed = pPlayer->GetFOV() < pPlayer->GetDefaultFOV();

	float flCycleTime = GetCycleTime();

	if ( bZoomed )
		flCycleTime = 0.135f;

	CSBaseGunFire( flCycleTime, m_weaponMode );
}
```

PR #699, opened 2021-10-28
API update SK #1.

StevenKal:
Adjust a few things in the API, then fix issue #498, then [this issue](https://dev-cs.ru/threads/1056/page-12#post-121492).
Check comments in the commits ("..." at the right of each) for detailled explanations.

PR #735, opened 2022-02-22
Fix player_weaponstrip

Vaqtincha:


StevenKal:
Tss, you say my #699 is "dumb" but you fix here something my PR already did, and in a best way, as it is more logic to "touch the finality code" for a better affect in overall. I feel some damn provocative moron style in this...

Vaqtincha:
A fix bil kriticheskim. jal.

PR #736, opened 2022-02-23
Fix teleport players to invalid spawn points

hajimura:
Resolves https://dev-cs.ru/threads/1056/post-121514

### DIFF
--- a/regamedll/dlls/triggers.cpp
+++ b/regamedll/dlls/triggers.cpp
@@ -1752,29 +1752,39 @@ void CBaseTrigger::TeleportTouch(CBaseEntity *pOther)
 		return;
 	}
 
-	Vector tmp = VARS(pentTarget)->origin;
-
-	if (pOther->IsPlayer())
+	if (IsSpawnPointValid(pOther, CBaseEntity::Instance(pentTarget))
 	{
-		// make origin adjustments in case the teleportee is a player. (origin in center, not at feet)
-		tmp.z -= pOther->pev->mins.z;
+		Vector tmp = VARS(pentTarget)->origin;
+
+		if (pOther->IsPlayer())
+		{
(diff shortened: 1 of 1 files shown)


RauliTop:
Will this cause a player in a server with semiclip can't teleport if another player at teleport spawn?

hajimura:
> Also, disabling mp_kill_filled_spawn makes all spawn points valids

So, i could add third bool parameter to IsSpawnPointValid func check or not this cvar

RauliTop:
What about first question?

StevenKal:
> What about first question?

Except when "mp_kill_filled_spawn" is "0", the "IsSpawnPointValid" simply searches if there is an entity in the sphere of the spawn, with 64 units max range, so, if another player is found, it will return false without taking care of the collisions properties set via semiclip plugins...
The modified code submitted simply prevents the teleport in case of the spawn "IsSpawnPointValid" returns false, rather than teleporting no matter what (so possibly on an invalid location [but should not be the case when destination's origin properly set], or, if there is already a client on the destination, as example).
But to handle better teleport on spawn & third-party plugins as semiclip, some other modifications will be required.
We could even start using "UTIL_EntitiesInBox" or even the unused "UTIL_IsSpawnPointOccupied" function, witch checks for boxes instead of sphere, a player has vertical rectangular boxes so it will make sense to use such, with a few modifications.
We could even add some kind of "UTIL_UnstuckClient" & use it at some places like for info_teleport_destination (hum, because two players who use the entry at the suite gonna spawn & do deep hugs!). But this code will prevent the second player to hug the first, till the first not moved around...

Vaqtincha:
Why?

hajimura:
Its unnecessary, because not gonna be merged 🚢


> Will this cause a player (in a server with semiclip) can't teleport if another player is at teleport spawn?

Yes

Issue #728, opened 2022-02-08
[feature request] upgrade swapteams command


rust2:
Could be added some option to `swapteams` command that will swap teams but also save team scores, frags and deaths?

rust2:
yes, completed ahahah

PR #739, opened 2022-03-19
Add an argument to `swapteams` command

ShadowsAdi:
Related to: https://github.com/s1lentq/ReGameDLL_CS/issues/728

### DIFF
--- a/regamedll/dlls/game.cpp
+++ b/regamedll/dlls/game.cpp
@@ -197,8 +197,17 @@ void GameDLL_EndRound_f()
 
 void GameDLL_SwapTeams_f()
 {
-	CSGameRules()->SwapAllPlayers();
-	CVAR_SET_FLOAT("sv_restartround", 1.0);
+	if (CMD_ARGC() == 1)
+	{
+		const char *pCmd = CMD_ARGV(1);
+		
+		CSGameRules()->SwapAllPlayers();
+		
+		if (pCmd[0] == '1' || !Q_stricmp(pCmd, "true"))
+		{
+			CVAR_SET_FLOAT("sv_restartround", 1.0);
+		}
+	}
 }
 
 #endif // REGAMEDLL_ADD
--- a/regamedll/dlls/game.cpp
+++ b/regamedll/dlls/game.cpp
@@ -197,7 +197,7 @@ void GameDLL_EndRound_f()
 
 void GameDLL_SwapTeams_f()
 {
-	if (CMD_ARGC() == 1)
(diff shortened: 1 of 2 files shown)


regamedll/dlls/game.cpp
```diff
@@ -197,8 +197,17 @@ void GameDLL_EndRound_f()
...
+	{
+		const char *pCmd = CMD_ARGV(1);
+		
+		CSGameRules()->SwapAllPlayers();
+		
+		if (pCmd[0] == '1' || !Q_stricmp(pCmd, "true"))
+		{
+			CVAR_SET_FLOAT("sv_restartround", 1.0);
+		}
+	}
```
wopox1337, line 210: now if you send swapteams without arguments, nothing will happen? Let's assume by default that the argument = 0

wopox1337:
@ShadowsAdi 

1) You broke the default behavior of the `swapteams` command (without arguments).
2) You don't need to make boolean notation (true\false) for CVars. It was always 1\0.

regamedll/dlls/game.cpp
```diff
@@ -197,8 +197,17 @@ void GameDLL_EndRound_f()
 
 void GameDLL_SwapTeams_f()
 {
-	CSGameRules()->SwapAllPlayers();
-	CVAR_SET_FLOAT("sv_restartround", 1.0);
+	if (CMD_ARGC() == 2)
```
wopox1337, line 200: You broke the default behavior of the swapteams command (without arguments).

regamedll/dlls/game.cpp
```diff
@@ -197,8 +197,17 @@ void GameDLL_EndRound_f()
...
+	{
+		const char *pCmd = CMD_ARGV(1);
+		
+		CSGameRules()->SwapAllPlayers();
+		
+		if (pCmd[0] == '1' || !Q_stricmp(pCmd, "true"))
```
wopox1337: You don't need to make boolean notation (true\false) for CVars. It was always 1\0.

wopox1337 (approved):

rust2:
What's the purpose of this change?

wopox1337:
> What's the purpose of this change and how does it relate to #728?

Ability to execute the command `swapteams`, swap commands and control the restart of the game (as well as not to use the restart).


from your issue:
> Could be added some option to `swapteams` command that **will swap teams but also save team scores, frags and deaths**?

rust2:
yes, very nice
so just swap teams but not force the round to restart???
then it was almost useless change
de_dust20000

PR #744, opened 2022-04-01
Expose findHook method

SmileYzn:
Expose findHook method

### DIFF
--- a/regamedll/public/regamedll/hookchains.h
+++ b/regamedll/public/regamedll/hookchains.h
@@ -67,6 +67,7 @@ public:
 
 	virtual void registerHook(hookfunc_t hook, int priority = HC_PRIORITY_DEFAULT) = 0;
 	virtual void unregisterHook(hookfunc_t hook) = 0;
+	virtual void findHook(hookfunc_t hook) = 0;
 };
 
 // Hook chain registry(for hooks [un]registration)
@@ -77,4 +78,5 @@ public:
 
 	virtual void registerHook(hookfunc_t hook, int priority = HC_PRIORITY_DEFAULT) = 0;
 	virtual void unregisterHook(hookfunc_t hook) = 0;
+	virtual void findHook(hookfunc_t hook) = 0;
 };
--- a/regamedll/regamedll/hookchains_impl.h
+++ b/regamedll/regamedll/hookchains_impl.h
@@ -205,6 +205,9 @@ public:
 	virtual void unregisterHook(hookfunc_t hook) {
 		removeHook((void*)hook);
 	}
+	virtual bool findHook(hookfunc_t hook) {
+		return findHook((void*)hook);
+	}
 };
 
 template <typename t_ret, typename t_class, typename ...t_args>
@@ -226,6 +229,9 @@ public:
 	virtual void unregisterHook(hookfunc_t hook) {
 		removeHook((void*)hook);
 	}
+	virtual bool findHook(hookfunc_t hook) {
+		return findHook((void*)hook);
+	}
 };
 
 template <typename t_ret, typename t_class, typename ...t_args>
@@ -246,5 +252,8 @@ public:
 	}
 	virtual void unregisterHook(hookfunc_t hook) {
 		removeHook((void*)hook);
+	}
+	virtual bool findHook(hookfunc_t hook) {
+		return findHook((void*)hook);
 	}
 };


SmileYzn:
My bad

ShadowsAdi:
What's the purpose of this?

SmileYzn:
> What's the purpose of this?

Expose findHook method to avoid double register that generate hlds error.
With this, i can check if already registered and skip register again during map changes.

I think ReHLDS need this too

ShadowsAdi:
1. ReAPI already does that: https://github.com/s1lentq/reapi/blob/master/reapi/src/hook_list.cpp#L276-L306
2. My suggestion is just to return if the chain is already hooked instead of stopping from execution: https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/regamedll/hookchains_impl.cpp#L55-L57

SmileYzn:
Who is telling about reapi? :P

I'm coding for metamod.

Need this to not implement a whole class only to check if a hook exists!

ShadowsAdi:
> Who is telling about reapi? :P
> 
> I'm coding for metamod.
> 
> Need this findHook to not implement a whole class only to check if a hook exists!

You should unregister all hooks on `Meta_Detach`, not just check if the hook already exists.

SmileYzn:
There is the problem, Meta_Death run only at server end not between change maps depending metamod plugin settings.
I'm using server activate / deactivate to this.

You just do not need to unregister all hooks only to not get this error. Need only to expose findHook method or just return instead of make a HLDS fatal error at registerHook Function.

Ps.
I'm trying to only register and unregister hooks at certain events in game, not hook it during all map execution.

wopox1337:
Why do you have the need to avoid Re-registering hooks again?

SmileYzn:
> Why do you have the need to avoid Re-registering hooks again?

I'm registering hooks according some events, not only just a registerHook at plugin begin and unregisterHook at plugin end.
So this change can easily fix that problems with re-registering.

Again, as [ShadowsAdi](https://github.com/ShadowsAdi) suggested, can only avoid Sys_Error call and replace with a simple return.

SmileYzn:
I'm fixed it 🙏

afwn90cj93201nixr2e1re:
.

afwn90cj93201nixr2e1re:
image

afwn90cj93201nixr2e1re:
.

SmileYzn:
I just try to prevent sys_error call. 
Again, Im doing my own metamod plugin, is not a reapi module or amxx plugin.

Soo I need to check if any hook is already registered with findHook method, or as you said. Make a return on it


I will post an example here

SmileYzn:
`#include "precompiled.h"

///////////////////////////////////////////////////////////////////////////////
// First case OK (This hook is running in while map is running)
///////////////////////////////////////////////////////////////////////////////

void DLL_POST_ServerActivate(edict_t* pEdictList, int edictCount, int clientMax)
{
	g_RehldsHookchains->ClientConnected()->registerHook(HOOK_ClientConnected);

	RETURN_META(MRES_IGNORED);
}

void DLL_POST_ServerDeactivate(void)
{
	g_RehldsHookchains->ClientConnected()->unregisterHook(HOOK_ClientConnected);

	RETURN_META(MRES_IGNORED);
}

///////////////////////////////////////////////////////////////////////////////
// Second case (This hook is running in while my mod is running)
//
// Game events can run in any moment, i can't prevent while will run in server
///////////////////////////////////////////////////////////////////////////////

void DLL_POST_ServerActivate(edict_t* pEdictList, int edictCount, int clientMax)
{
	g_RehldsHookchains->ClientConnected()->registerHook(HOOK_ClientConnected);

	RETURN_META(MRES_IGNORED);
}

// At any event what i want i need to register hook
void GAME_FirstEvent()
{
	// Now i already registered event, cant need to load twice.
	// Or just prevent Sys_Error call in registerHook while event will be registered
	if(g_RehldsHookchains->ClientConnected()->findHook(HOOK_ClientConnected) == false)
	{
		g_RehldsHookchains->ClientConnected()->registerHook(HOOK_ClientConnected);
	}
}

// At any event what i want i need to un-register hook
void GAME_SecondEvent()
{
	// Do not need to use find hook, Sys_Error call will not run at unregisterHook
	g_RehldsHookchains->ClientConnected()->unregisterHook(HOOK_ClientConnected);
}

void DLL_POST_ServerDeactivate(void)
{
	// Do not need to use find hook, Sys_Error call will not run at unregisterHook
	g_RehldsHookchains->ClientConnected()->unregisterHook(HOOK_ClientConnected);

	RETURN_META(MRES_IGNORED);
}`

Remember: I a case that i can't prevent when my mod game event will be load or unloaded

afwn90cj93201nixr2e1re:
.

afwn90cj93201nixr2e1re:
.

SmileYzn:
> Are you trying to implement PUG mode as meta plugin by any chance?

Yes, this is a better example what I trying to do

afwn90cj93201nixr2e1re:
.

SmileYzn:
You put a return in function, just still hooked. I want to really  unregister chain. The sys error call do not make any sense while is already registered. Need only to return in code.

And avoid unnecessary code to manage hooks.

Amxmodx does that with some modules like fakemeta

afwn90cj93201nixr2e1re:
.

SmileYzn:
There is, unregister hook before register a new one, that do not make sense, just to avoid sys_error.

afwn90cj93201nixr2e1re:
.

SmileYzn:
Only a returns is ok to me, since is really what i need.

My point if view is Sys_Error is not really needed if you won't take any action at registerHook that can crash the server. 
Can show the error in server console or something else, in really is not needed to crash server anyway.

PR #751, opened 2022-04-09
Clamp moving entities' sounds volume

etojuice:


StevenKal:
That PR had an impressive very short lifetime!

PR #754, opened 2022-04-28
Implement `game_round_start` & `game_entity_restart` triggers

ShadowsAdi:
https://github.com/ValveSoftware/halflife/issues/2190

### DIFF
--- a/regamedll/dlls/multiplay_gamerules.cpp
+++ b/regamedll/dlls/multiplay_gamerules.cpp
@@ -2029,7 +2029,9 @@ void EXT_FUNC CHalfLifeMultiplay::__API_HOOK(RestartRound)()
 
 			pPlayer->RoundRespawn();
 			
+#ifdef REGAMEDLL_ADD
 			FireTargets("game_round_start", pPlayer, nullptr, USE_TOGGLE, 0.0);
+#endif
 		}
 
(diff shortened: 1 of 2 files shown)


s1lentq:
it makes sense to use `game_entity_restart` instead it `game_round_start`,
because `game_round_start` can be used to start once on actually game start round

regamedll/dlls/multiplay_gamerules.cpp
```diff
@@ -2066,6 +2070,8 @@ void EXT_FUNC CHalfLifeMultiplay::__API_HOOK(RestartRound)()
 	m_bTargetBombed = m_bBombDefused = false;
 	m_bLevelInitialized = false;
 	m_bCompleteReset = false;
+	
+	FireTargets("game_round_start", nullptr, nullptr, USE_TOGGLE, 0.0);
```
wopox1337, line 2074: Why this without macro?

```suggestion

#ifdef REGAMEDLL_ADD
	FireTargets("game_round_start", nullptr, nullptr, USE_TOGGLE, 0.0);
#endif
```

ShadowsAdi: You're right, thank you!

PR #755, opened 2022-04-29
Fix long standing jump bug & movement lag based on FPS

ShadowsAdi:
```cpp
// Legacy movement. The new one fixes FPS based jump and movement.
// 0 - Legacy behavior
// 1 - New behavior
//
// Default value: "0"
sv_legacy_movement 0
```

~**IT SHOULD BE TESTED MORE, I TESTED IT ONLY IN SINGLE PLAYER**~

https://github.com/ValveSoftware/halflife/issues/81
https://github.com/ValveSoftware/halflife/issues/682

https://github.com/pmrowla/hl2sdk-csgo/blob/master/game/shared/gamemovement.cpp#L1211

I'm sorry for the build.yml changes, just tested a few ways of achieving this.


RauliTop:
@ShadowsAdi are you already test this on multiplayer? Working well?

This will change the behaviour of gameplay, probably a cvar to disable should be great.

ShadowsAdi:
> have you already test this on multiplayer? Working well?

Yes, it's working just fine on multiplayer too, on a 32/32 server.

> This will change the behaviour of gameplay, probably a cvar to disable should be great.

I don't think it requires a cvar because it fixes something that is't kinda broken. That slow was just annoying while using lower or higher FPS than 100.

RauliTop:
> > have you already test this on multiplayer? Working well?
> 
> Yes, it's working just fine on multiplayer too, on a 32/32 server.
> 
> > This will change the behaviour of gameplay, probably a cvar to disable should be great.
> 
> I don't think it requires a cvar because it fixes something that it's kinda broken. That slow was just annoying while using lower or higher FPS than 100.

Physics are dependant from FPS, yes, it's a bug.

But some mods use this logic, for example KZ or HNS. If you have less fps, like 20, your slowdown jump is less than having 80-100 fps.

Consider adding the cvar, please. It's not for me, I'm talking about other players

ShadowsAdi:
> But some mods use this logic, for example KZ or HNS. If you have less fps, like 20, your slowdown jump is less than having 80-100 fps.

This fix should help especially these mods because you can't gain more speed while bhopping ( as before ) it just fix the movement slow after jumping.

> Consider adding the cvar, please. It's not for me, I'm talking about other players

I still don't see the relevance of adding a cvar just to fix something that was broken. This implementation is still under `REGAMEDLL_FIXES` macro so it's behaviour can be always changed as it was before.

fred0r:
been using this now for about ~1,5 weeks now and it seem to behave correct.
a cvar would indeed be nice to make it easier to check the changed behavior, since there would be no need to stop/replace the cs.so every time.

Garey27:
There will be same behaviour no matter frametime * 1000 or msec.
https://github.com/s1lentq/ReGameDLL_CS/blob/b9cccc691bdabbf9cb573be8ee5e39c9a4f70c4a/regamedll/pm_shared/pm_shared.cpp#L2841
So tell me what a difference between this logic:
- reduce timers by cmd.msec
or
- reduce timer by frametime which are calulcated also by cmd.msec ? This PR is useless.

Garey27:
Also if u look in csgo CGameMovement::ReduceTimers there global frametime not players one. Which is wrong for Half-Life/CS because we simulate shared movement code where timers reduces by player fps! not server one.

ShadowsAdi:
https://github.com/s1lentq/ReGameDLL_CS/blob/b9cccc691bdabbf9cb573be8ee5e39c9a4f70c4a/regamedll/pm_shared/pm_shared.cpp#L2841

I know, but it seems working correctly as it is now.

> There will be same behaviour no matter frametime * 1000 or msec.

> * This PR is useless.

Test it out, compare the results by yourself and reply to this thread again.

Garey27:
Ok i was tested and its absolutely same behaviour as without fix. Notice the slowdown with 500 fps and no slowdown with 20 fps. (jump crouched > unduck when touch ground and hold +forward). 
There no difference.
https://user-images.githubusercontent.com/4866813/169670327-1272b3f0-8c08-4bdd-a645-4b13862741a4.mp4

.

ShadowsAdi:
> Ok i was tested and its absolutely same behaviour as without fix. Notice the slowdown with 500 fps and no slowdown with 20 fps. (jump crouched > unduck when touch ground and hold +forward). There no difference. https://user-images.githubusercontent.com/4866813/169670327-1272b3f0-8c08-4bdd-a645-4b13862741a4.mp4
> 
> .

After some debuggings, I found out that the windows version of the ReGameDLL is not based on any of the additional preprocessor defs. It is just a pure version without any additional fixes.

In linux, it seems to build it using the preprocessor definitions.

https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/msvc/ReGameDLL.vcxproj#L959

fred0r:
pls approve workflow

lespaul64:
Finally someone willing to fix this issue. This is a game bug, fix is a must.

Yes i noticed this bug too especially when playing fps more than 100~120. The only way to get fast jump penalty is to use fps_max 60, 60fps to play fps game haha this is terrible. With 60fps i can't even climb a pyramid, but with 256fps i can easily climb but this jump penalty bug is annoying. In certain mode for example ZE, we all know this mode is base on speed, but with so many obstacle that need to jump and duck i lost because of jump penalty.

I play csgo too, but csgo doesn't have this bug, i can play more than 300fps.

Where is the compiled binary for this latest PR? can't even find it. I wanna test too.

ShadowsAdi:
Until this workflow is approved, you can download the compiled binaries for
this PR here:
https://github.com/ShadowsAdi/ReGameDLL_CS/actions/runs/2413788290

În mar., 31 mai 2022 la 10:17, lespaul64 ***@***.***> a scris:

ShadowsAdi:
@lespaul64 could you test it after changes and review it?

RauliTop:
Please @ShadowsAdi, explain that new magical numbers

lespaul64:
> @lespaul64 could you test it after changes and review it?

Ok the new patch is acceptable for me, can feel the different. Playing with 256fps now feel a bit similar to 60fps jump penalty behavior. Maybe need so tweaks on the magic numbers & calculations.

ShadowsAdi:
> Please @ShadowsAdi, explain that new magical numbers

> Ok the new patch is acceptable for me, can feel the different. Playing with 256fps now feel a bit similar to 60fps jump penalty behavior. Maybe need so tweaks on the magic numbers & calculations. BTW where did you get that numbers?

I tested some values locally, on a windows build running at 1000 FPS + client running at 700-900 FPS.
These values seemed to fix the issue, I will do some more math to gain better results.

https://github.com/s1lentq/ReGameDLL_CS/pull/755/files#diff-2aeae34eebf30e19694d07e23caecbe3a34e1d0fdc4cec2fd75fd14b34a1c10bR2781 

From my tests, `cmd.msec` values will be greater than 10 while client is running at < 150 FPS and will be lower than 10 from 150 FPS above.

ShadowsAdi:
> Sorry for noob question, but why not put a cvar to enable/disable this FPS fix with cvars to control these values if fix is enabled on server?

Because that value should not be modified, same as other important values in gamedll.

dist/game.cfg
```diff
@@ -483,3 +483,10 @@ sv_enablebunnyhopping 0
 //
 // Default value: "0"
 mp_plant_c4_anywhere 0
+
+// Legacy movement. The new one fixes FPS based jump and movement.
+// 0 - Legacy behavior
+// 1 - New behavior
+//
+// Default value: "0"
+sv_legacy_movement 0
```
wopox1337, line 492:
```suggestion
sv_legacy_movement 0

```

regamedll/dlls/game.cpp
```diff
@@ -162,6 +162,7 @@ cvar_t allchat                           = { "sv_allchat", "0", 0, 0.0f, nullptr
 cvar_t sv_autobunnyhopping               = { "sv_autobunnyhopping", "0", 0, 0.0f, nullptr };
 cvar_t sv_enablebunnyhopping             = { "sv_enablebunnyhopping", "0", 0, 0.0f, nullptr };
 cvar_t plant_c4_anywhere                 = { "mp_plant_c4_anywhere", "0", 0, 0.0f, nullptr };
+cvar_t sv_legacy_movement                = { "sv_legacy_movement", "0", FCVAR_SERVER, 0.0f, nullptr };
```
wopox1337, line 165:
```suggestion
cvar_t sv_legacy_movement                = { "sv_legacy_movement", "0", 0, 0.0f, nullptr };
```

wopox1337 (changes requested):

dist/game.cfg
ShadowsAdi: I don't understand this change.

SmileYzn:
> > Sorry for noob question, but why not put a cvar to enable/disable this FPS fix with cvars to control these values if fix is enabled on server?
> 
> Because that value should not be modified, same as other important values in gamedll.

In your tests you changed values, OK

In my tests i can't change values? why?

I did not understand, ReHLDS ecosystem has a lot of these 'magical' values by cvars. What is the problem?

I really understand that variables and values are really important, but the reason to give the cvars for change according cvars that do not make sense to me.

Also other thing, is really a BUG because the original HLDS do not have values fixed and other small changes in code. Bug let's threat this as feature than a bug is not a better way in this case?

ShadowsAdi:
> In your tests you changed values, OK
> 
> In my tests i can't change values? why?
> 
> I did not understand, ReHLDS ecosystem has a lot of these 'magical' values by cvars. What is the problem?

Right now, this is reGameDLL, it hasn't "magical" values by cvars when you shoot with a weapon, or when you reload one. They are hardcoded values for a reason.

> Also other thing, is really a BUG because the original HLDS do not have values fixed and other small changes in code. Bug let's threat this as feature than a bug is not a better way in this case?

If more people complain for that, it's not a feature.

ShadowsAdi:
> I forgot to mention, in first release of this fps fix, the step sounds is a bit accelerated in remote server with pingboost 3, sys_ticrate 1000 and rate 100000 and updaterate & cmdrate 102.

What do you mean by that?

SmileYzn:
Sorry for bump, but i want to test these changes.
Some news? 

Thanks :+1:

SmileYzn:
> @SmileYzn You can test with artifact releases monocle_face

Yes but it has requested changes, anyway i think will not change anything with

ShadowsAdi:
> > @SmileYzn You can test with artifact releases monocle_face
> 
> Yes but it has requested changes, anyway i think will not change anything with

It shows that because it hasn't been reviewed yet. I added the suggestions proposed by @wopox1337.

fred0r:
```
5.21.0.573-dev on linux 
-pingboost 3
-sys_ticrate 1016
schedtool with SCHED_RR
```
players complain that they can jump further/higher than before - running at 100 fps.

wopox1337 (changes requested):
A better (detailed) description of at least the new behavior is needed.

README.md
```diff
@@ -110,6 +110,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_hostages_rescued_ratio          | 1.0     | 0.0 | 1.0          | Ratio of hostages rescued to win the round. |
 | mp_legacy_vehicle_block            | 1       | 0   | 1            | Legacy func_vehicle behavior when blocked by another entity.<br/>`0` New behavior <br/>`1` Legacy behavior |
 | mp_dying_time                      | 3.0     | 0.0 | -            | Time for switch to free observing after death.<br/>`0` - disable spectating around death.<br/>`>0.00001` - time delay to start spectate.<br/>`NOTE`: The countdown starts when the player’s death animation is finished.|
+| sv_legacy_movement                 | 0       | 0   | 1            | Legacy movement. The new one fixes FPS based jump and movement.<br/>`0` Legacy behaviour <br/>`1` New behaviour |
```
wopox1337, line 113: A better (detailed) description of at least the new behavior is needed.

SmilexGamer:
Will this request be pushed soon?

PR #782, opened 2022-10-12
New Cvar: `mp_drop_grenade_enable`

JulioBarker:
Allow players to drop grenades from their inventory

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -7930,7 +7930,7 @@ CBaseEntity *EXT_FUNC CBasePlayer::__API_HOOK(DropPlayerItem)(const char *pszIte
 	{
 		if (!pWeapon->CanDrop()
 #ifdef REGAMEDLL_ADD
-		|| (!drop_grenade_enable.value && IsGrenadeWeapon(pWeapon->m_iId))
+		|| (!drop_grenade_enable.value && IsGrenadeWeapon(pWeapon->m_iId)) || (IsGrenadeWeapon(pWeapon->m_iId) && m_rgAmmo[pWeapon->PrimaryAmmoIndex()] <= 0)
 #endif
 		)
 		{
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -7928,7 +7928,11 @@ CBaseEntity *EXT_FUNC CBasePlayer::__API_HOOK(DropPlayerItem)(const char *pszIte
 #endif
 	if (pWeapon)
 	{
-		if (!pWeapon->CanDrop())
+		if (!pWeapon->CanDrop()
+#ifdef REGAMEDLL_ADD
+		|| (!drop_grenade_enable.value && IsGrenadeWeapon(pWeapon->m_iId))
+#endif
+		)
 		{
 			ClientPrint(pev, HUD_PRINTCENTER, "#Weapon_Cannot_Be_Dropped");
 			return nullptr;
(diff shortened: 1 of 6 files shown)


StevenKal:
I might recommend to put a CVar via flags in order to control the manual drop "per grenade", then put the CVar code inside each function "*::CanDrop".
Like I did for a module [here](https://www.amxmod.net/forum/showthread.php?tid=1583), so, more customizable.

Vaqtincha:
regame is slowly turning into a circus

wopox1337 (changes requested):
Found bug

Desc: It is enough to press `G` after throwing this grenade, but before deploy new weapon. It will drop fake nade

https://www.youtube.com/watch?v=aE1eI4Yud1I

Vaqtincha:
Типичный шорох: мержит все что не проверено и бесполезное и не мержит правки которые тестировали 2-3 года.

wopox1337:
@JulioBarker need changes

Vaqtincha:
then players will start spamming with grenades. and you will add another million cvars to control it. 

IMHO it's better to do this with 3rd party plugins.

Vaqtincha:
What else will happen if I have 2 flash? Will drop everything. Although it should be one at a time, but this requires an additional code.

Vaqtincha:
Шолох давай сюда умник.

wopox1337 (changes requested):
still buggy with 2 flashes

dystopm:
IMO, this should be a 3rd party plugin. Idea is good, but:
1) Dropping grenade per grenade would spam world with entities (considering you can extend grenade carry limit), leading to (2)
2) If you have ammo > 1, then you'll drop a single grenade weaponbox containing +2 ammo on it, which is aesthetically uncomfortable, leading to (3)
3) Use a different model (ammo box maybe? something like models/w_argrenade.mdl) for grenade weaponboxes but only when ammo > 1, but that will turn this to an extra feature, and an extra model precaching, leading to go outside the limits of the purpose of this project.

q52888940:
Why i can drop knife?????

regamedll/dlls/player.cpp
```diff
@@ -7928,7 +7928,11 @@ CBaseEntity *EXT_FUNC CBasePlayer::__API_HOOK(DropPlayerItem)(const char *pszIte
 #endif
 	if (pWeapon)
 	{
-		if (!pWeapon->CanDrop())
+		if (!pWeapon->CanDrop()
+#ifdef REGAMEDLL_ADD
+		|| (!drop_grenade_enable.value && IsGrenadeWeapon(pWeapon->m_iId)) || (IsGrenadeWeapon(pWeapon->m_iId) && m_rgAmmo[pWeapon->PrimaryAmmoIndex()] <= 0)
+#endif
```
wopox1337, line 7934: Check how `mp_nadedrops` works and redo the code to take into account the `weaponbox` requirements and player checks that are already there.

wopox1337 (changes requested):
check that https://github.com/s1lentq/ReGameDLL_CS/blob/c48be874743b2a440728889acb4797f4ec04137a/regamedll/dlls/player.cpp#L1578-L1589

Vaqtincha:
Бесполезно. Зачем вообще бросать гранату на землю? Чтобы засорять карту флудом?

StevenKal:
And why talk in Russian when you can do efforts to talk in English!

> Бесполезно. Зачем вообще бросать гранату на землю? Чтобы засорять карту флудом?

And why talk in Russian when you can do efforts to talk in English! Still same shits with you!

Vaqtincha:
> > Бесполезно. Зачем вообще бросать гранату на землю? Чтобы засорять карту флудом?
> 
> And why talk in Russian when you can do efforts to talk in English! Still same shits with you!

Я тебя не ответил русофоб. Иди мимо пожалуйста.

PR #786, opened 2022-11-21
New CVar: `mp_hostages_rescued_ratio`

fl0werD:
https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/dlls/multiplay_gamerules.cpp#L1471
This business logic does not work correctly because the code below does not consider by 50%.
https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/dlls/multiplay_gamerules.cpp#L1482

Adding cvar for ratio and and default value for compatibility.


wopox1337:
:D
The standard conditions of the game are quite strange

https://youtu.be/lvXW_PIqYN0

RauliTop:
@fl0werD @wopox1337 
Also, we must disable bHostageAlive var with #ifndef REGAMEDLL_ADD

fl0werD:
@RauliTop it able to be more ifdef macros, but it unreadable fine. We can also wrap more code into macros, but now it's minimal edit.

PR #788, opened 2022-11-30
fix: Reset immunity effects always

Vaqtincha:


RauliTop:
Why? How about if are we using a plugin to add these effects?
and why you set the STATUSICON_HIDE out of the condition?

Cvar check is a must.

Vaqtincha:
> Cvar check is a must.
No!

set mp_respawn_immunitytime to a higher value (e.g. 10) and after the player spawns immediately set mp_respawn_immunity_effects to 0 and he will use this effect throughout the map after the end of the protection.

As for the status icon, yes I agree (but this is not my cant)

Vaqtincha:
> m_bProtectionEffects

There will be no benefit from this member for api. Better to use iuser3 (like PLAYER_IN_BOMB_ZONE)

wopox1337:
> > m_bProtectionEffects
> 
> There will be no benefit from this member for api.
> 
> Better to use iuser3 ? (like PLAYER_IN_BOMB_ZONE)

Using a network member is not the best solution.

wopox1337 (approved):

PR #811, opened 2023-02-21
fix swapteams not saving player/team scores (sv_restartround logic)

deprale:
swapteams calls sv_restartround 1 which will reset all team & player scores, total rounds played etc, this PR should fix this unintended behaviour.

the best way i could do it is by adding a global boolean.

(sorry for new PR i couldn't git reset --hard the last one I closed because I deleted it before I went to sleep)

Also I tested this properly with swapteams -1, swapteams with no arg, and swapteams with different values, and it worked as intended with the PR. Also ran a workflow for this on my own repo and it passed [all tests](https://github.com/deprale/ReGameDLL_CS/actions/runs/4228715979).

### DIFF
--- a/regamedll/dlls/game.cpp
+++ b/regamedll/dlls/game.cpp
@@ -212,6 +212,7 @@ void GameDLL_SwapTeams_f()
 
 	if (value > 0.0f)
 	{
+		g_bSwappedTeams = true;
 		CVAR_SET_FLOAT("sv_restartround", value);
 	}
 }
--- a/regamedll/dlls/globals.cpp
+++ b/regamedll/dlls/globals.cpp
@@ -12,3 +12,4 @@ bool g_bIsBeta = false;
 bool g_bIsCzeroGame = false;
 bool g_bAllowedCSBot = false;
 bool g_bHostageImprov = false;
+bool g_bSwappedTeams = false;
--- a/regamedll/dlls/globals.h
+++ b/regamedll/dlls/globals.h
@@ -39,3 +39,4 @@ extern bool g_bIsBeta;
 extern bool g_bIsCzeroGame;
 extern bool g_bAllowedCSBot;
 extern bool g_bHostageImprov;
+extern bool g_bSwappedTeams;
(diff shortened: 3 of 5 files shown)


StevenKal:
I might suggest instead, to change behavior of CVar "sv_restartround", in order to make a "current round restart" (CHalfLifeMultiplay 's "m_bCompleteReset" not set), then let CVar "sv_restart" do a complete one as now, which is feasible.

Then for the command "swapteams", make a format like:
swapteams [complete reset (0|1)] - Instantly swap the teams.
So, we could type:
swapteams -> To swap teams, nothing else.
swapteams 0 -> To swap teams, and restart only current round (not the scores & weapons, etc.).
swapteams 1 -> To swap teams, and restart all the rounds as usual (scores, weapons, etc.).

Because there are two CVars for restarting rounds, but both do the same, which is a bit stupid.
My suggestion will add a new feature, the only "inconvenient" is that admins will have to change their binds (or/and AMX[X]'s command lines & config files), from "sv_restartround" to "sv_restart", to make that work as before (complete reset of scores & cie).
But they could still add two lines in the menus.

deprale:
> I might suggest instead, to change behavior of CVar "sv_restartround", in order to make a "current round restart" (CHalfLifeMultiplay 's "m_bCompleteReset" not set), then let CVar "sv_restart" do a complete one as now, which is feasible.

It won't work this way, since m_bCompleteReset calls pPlayer->Reset(); that resets the current iAccount values and resets their inventories (weapons nades etc) as if they just spawned for the right team.

SmileYzn:
endround?

deprale:
> endround command for this logic ??

Wouldn't that increase the m_iTotalRoundsPlayed as well? Then you have to do 3 things just to swap the teams, yes arguably you could also do swapteams -1 at pre newround but then you'll also have to strip all user weapons and then rg_give_default_items, which all seem like kinda stupid solutions...

The whole point of swapteams is that it should keep the score regardless if the game is restarted or not.

PM32:
Hi @deprale , considering the "swapteams" command is an added feature, you should consider using REGAMEDLL_ADD macro on every conditional using g_bSwappedTeams, including its declaration.

wopox1337:
> swapteams calls sv_restartround 1 which will reset all team & player scores, total rounds played etc...

Just use `swapteams 0` - to swap teams without restart round and resetting all.
https://github.com/s1lentq/ReGameDLL_CS/blob/e1d1c11992e6f97037f6e6ab5baba015d2922db0/regamedll/dlls/game.cpp#L215-L218

@deprale If I misunderstood your intentions, please let me know in this PR.

PR #815, opened 2023-03-20
Fix incorrect 3rd camera player death animations when frozen.

deprale:
Player dies with FL_FROZEN flag set

https://user-images.githubusercontent.com/33827235/226233916-bc808483-1076-4416-aab3-dc0a77fdca9b.mp4

fixed

https://user-images.githubusercontent.com/33827235/226233922-e759a377-bdb5-4a59-83d9-6033c71fb831.mp4

edit: 
granted, this can also be fixed with amxx itself, just reset FL_FROZEN on pre playerkilled, and there's no default weapon that freezes you (at least in cstrike), and a good coder would just remove that flag on death, but maybe it would be better to take care of that straight from regamedll?
Can map triggers set this flag to players? If yes, wouldn't it make sense to remove this flag on player killed?

Please share your thoughts.


RauliTop:
I realized it some years ago, and I fixed by the way you said in amxx.

but, of course, it will be a good addition to have it by default on regamedll
many coders don't know it

PR #822, opened 2023-04-06
de_dust2 Rampa Bug Speed

scripting6:
Hi,
I was playing counter strike 1.6 from 2005 and when have FreezeTime Terrorists can do double ducks when they on rampa
and move towards the long (In Speed Like 200 - sv_maxspeed 200)
and now when I play with all updates to CS, ReHLDS, ReGameDLL and more. I do double duck and I have 2 problems.

first, Ducks Rampa Bug - dont work when player use fps_override 1
second, this bug is too 'slowly' ducks and I want to fastest the ducks maybe cvar - sv_maxspeed_freezetime_ducks

will help to cs community, maybe another servers want to close double ducks bug.. so they will use sv_maxspeed_freezetime_ducks 0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,8 +20,6 @@ jobs:
       solution: 'msvc/ReGameDLL.sln'
       buildPlatform: 'Win32'
       buildRelease: 'Release'
-      buildReleasePlay: 'Release Play'
-      buildTests: 'Tests'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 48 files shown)


wopox1337:
I may have understood your problem, but a more detailed description is required. 

Do you want to fix the bug? Moreover, you should go to the Issue section, not Pull Request.

https://www.youtube.com/watch?v=4T7w2bnenBg

PR #831, opened 2023-05-16
Little code cleaning: g_vecAttackDir

dystopm:
Same argument as #830 : variables used as parameters, mostly used in a old fashioned programming way and on singleplayer hlsdk branch.

Might be useless but opens mind about certain code that should be reviewed in detail.


dystopm:
> There's no need to wrap in a macro `REGAMEDLL_FIXES` for each change, except when the behavior does change, so in this case it's just a refactoring and I would prefer not to add this the macro.
> 
> p.s We shouldn't be concerned about the fact that with these minor changes we can break compatibility 3-rd party things when they use memhacks.

Thanks for your observation, I'll keep in mind this next time 🙌 commit done now

PR #832, opened 2023-05-16
Fix Glock and Famas undesired ammo decreasing on burst mode

PM32:
Consider the next scenario: 

1) Get a glock with clip-bpammo of 20-1
2) Shoot 3 bullets on burst mode, leaving clip-bpammo of 17-1
3) Try to reload, won't let

You need to call TabulateAmmo to enable reload again (shoot in single mode).

There is no reason to decrease bpammo on clip decreasing action anyways.
(and there is no reason to use ammo_ members from CBaseEntity)

### DIFF
--- a/regamedll/dlls/weapons.cpp
+++ b/regamedll/dlls/weapons.cpp
@@ -761,8 +761,9 @@ void CBasePlayerWeapon::FireRemaining(int &shotsFired, float &shootTime, BOOL bI
 	if (bIsGlock)
 	{
 		vecDir = m_pPlayer->FireBullets3(vecSrc, gpGlobals->v_forward, 0.05, 8192, 1, BULLET_PLAYER_9MM, 18, 0.9, m_pPlayer->pev, true, m_pPlayer->random_seed);
+#ifndef REGAMEDLL_FIXES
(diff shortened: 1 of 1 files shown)


dystopm:
@s1lentq @wopox1337 it's worth it?

regamedll/dlls/weapons.cpp
```diff
@@ -761,16 +761,19 @@ void CBasePlayerWeapon::FireRemaining(int &shotsFired, float &shootTime, BOOL bI
 	if (bIsGlock)
 	{
 		vecDir = m_pPlayer->FireBullets3(vecSrc, gpGlobals->v_forward, 0.05, 8192, 1, BULLET_PLAYER_9MM, 18, 0.9, m_pPlayer->pev, true, m_pPlayer->random_seed);
+#ifndef REGAMEDLL_FIXES
 		--m_pPlayer->ammo_9mm;
```
wopox1337, line 765: I can't understand the original logic behind valve's original GameDLL. Why does it need to reduce ammo from the holster every time a shot is fired? It's weird.

wopox1337 (approved):

PR #844, opened 2023-07-13
Fixed crash sometimes occurring while map analyzing for zbot navigation

s1lentq:
The iterator `m_analyzeIter` could potentially become invalid due to the possibility of `TheNavAreaList` removing elements while the analysis is ongoing.
It's recommended to test the proposed fix thoroughly before merging it


dystopm:
Good catch

PR #851, opened 2023-07-17
`API`: CSPlayer new members (physics related)

dystopm:
- Added 4 members to CSPlayer class:
     - `m_flJumpHeight`: player vertical jump force, 0 = default (~268.3)
     - `m_flLongJumpHeight`: player vertical jump force with longjump enabled, 0 = default (~299.3)
     - `m_flLongJumpForce`: player horizontal jump force with longjump enabled, 0 = default (560.0)
     - `m_flDuckSpeedMultiplier`: player duck speed multiplier, 0 = default (0.3333)
- Cached pmove player's CSPlayer pointer in a global variable

### DIFF
--- a/regamedll/pm_shared/pm_shared.cpp
+++ b/regamedll/pm_shared/pm_shared.cpp
@@ -12,6 +12,10 @@ char pm_grgchTextureType[MAX_TEXTURES];
 playermove_t *pmove = nullptr;
 BOOL g_onladder = FALSE;
 
+#ifdef REGAMEDLL_API
+static CCSPlayer *pmoveplayer = nullptr;
+#endif
+
 #ifdef CLIENT_DLL
 	int iJumpSpectator;
 	float vJumpOrigin[3];
@@ -1902,10 +1906,8 @@ void PM_Duck()
 	real_t mult = PLAYER_DUCKING_MULTIPLIER;
 
 #ifdef REGAMEDLL_API
-	const CCSPlayer* player = UTIL_PlayerByIndex(pmove->player_index + 1)->CSPlayer();
-
-	if (player->m_flDuckSpeedMultiplier > 0.0)
-		mult = player->m_flDuckSpeedMultiplier;
+	if (pmoveplayer->m_flDuckSpeedMultiplier > 0.0)
+		mult = pmoveplayer->m_flDuckSpeedMultiplier;
 #endif
 
 	pmove->cmd.forwardmove *= mult;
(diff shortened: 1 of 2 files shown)


dystopm:
@wopox1337 @s1lentq need a hand here, any clue of why test demos are failing with this?

s1lentq:
@dystopm 

Fields like `clientdata_t`, `entity_state_t`, etc. are transmitted over the network, so input data of these fields are registered in the demo through `sendto`.
There is a possibility that differences between `double` and `float` with a difference of `>0.0001` lead to fail the demo test because output data is not equal to input data.
Thus the demo file was recorded in completely vanilla code and `PLAYER_DUCKING MULTIPLIER` explicitly uses double-precision floatpoint (without `.f`), but in this PR use float type.
In practice, this difference is insignificant, but it is important to maintain it for the purpose of the demo test.

dystopm:
> @dystopm
> 
> Fields in structs `clientdata_t`, `entity_state_t`, etc. are transmitted over the network, so input data of these fields are registered in the demo through `sendto`. There is a possibility that differences between `double` and `float` with a difference of `>0.0001` lead to fail the demo test because output data is not equal to input data. Thus the demo file was recorded in completely vanilla code and `PLAYER_DUCKING MULTIPLIER` explicitly uses double-precision floatpoint (without `.f`), but in this PR use float type. In practice, this difference is insignificant, but it is important to maintain it for the purpose of the demo test.

That was clever. I thought for a moment of that kind of issue guessing about a precision issue but I just stared at the fact all three memory types were float, I totally dismissed `PLAYER_DUCKING_MULTIPLIER` is a double for no having `.f` suffix. Thanks again! 👌🏼

regamedll/pm_shared/pm_shared.cpp
```diff
@@ -2465,6 +2483,24 @@ void PM_Jump()
 	{
 		PM_PlayStepSound(PM_MapTextureTypeStepType(pmove->chtexturetype), fvol);
 	}
+	
+#ifdef REGAMEDLL_API
+	auto PM_JumpHeight = [&player](bool longjump) -> real_t
+#else 
+	auto PM_JumpHeight = [&](bool longjump) -> real_t
```
s1lentq, line 2490: Using a lambda inside a function may not be a good practice.
It can make the function bulky and reduce code readability. It might be better to consider using the inline function separately instead.

p.s
It should be noted that there are places in the project code where lambdas are used, but at that time they were not given too much attention to discuss.

regamedll/pm_shared/pm_shared.cpp
```diff
@@ -2015,7 +2024,16 @@ void EXT_FUNC __API_HOOK(PM_LadderMove)(physent_t *pLadder)
 
 		if (pmove->flags & FL_DUCKING)
 		{
-			flSpeed *= PLAYER_DUCKING_MULTIPLIER;
+			real_t mult = PLAYER_DUCKING_MULTIPLIER;
+
+#ifdef REGAMEDLL_API
+			const CCSPlayer* player = UTIL_PlayerByIndex(pmove->player_index + 1)->CSPlayer();
```
s1lentq, line 2030: It would be better to get pointer once in `PM_Move`.
This would be a good approach if it could be ensure that `player_index` will not change from the API side throughout the entire `PM_Move` call.

dystopm: What about using a global CCSPlayer variable in pm_shared.cpp ?

s1lentq: I would prefer to do it directly in `pmove`, but since the engine is responsible for forming this pointer with the structure, there is no possibility to do so. Therefore, using a global variable is the only option available in this case.

You can use something like:
`static CCSPlayer *pmoveplr = NULL;`

regamedll/pm_shared/pm_shared.cpp
dystopm: Done! Check commit

regamedll/pm_shared/pm_shared.cpp
dystopm: Done! Check commit

s1lentq (approved):

PR #853, opened 2023-07-22
Ensure m_pDriver assignation on func_vehicle only

dystopm:
`m_pDriver` is a member present only on `CFuncVehicle` class, in `CBasePlayer` workflow there's a train quitting logic where game is setting `m_pDriver` to `NULL` on both entities **func_vehicle** and **func_tracktrain**, which shares `FCAP_DIRECTIONAL_USE` ObjectCaps flag but the last one doesn't have m_pDriver member (including the implicit cast to CFuncVehicle on code)

Workaround (and ToDo): Train quitting logic is quite similar, should that be grouped in a single function?


wopox1337 (approved):

wopox1337:
>  Train quitting logic is quite similar, should that be grouped in a single function?

@dystopm sure

PR #861, opened 2023-09-11
`FIX`: Reloading animation bug while holding weapon with altered Maxclip

dystopm:
Sort of an old bug. If client prediction is activated (`cl_lw 1`), player is holding a weapon whose maxclip was altered (easily nowadays with `rg_set_iteminfo` natives), active weapon's clip is equal to altered maxclip, and then presses the reload button, it will reproduce a client-side reload animation which stops until player releases the button, because client thinks you met the conditions to reload the weapon considering its clip differs from its maxclip.

This fixes that weapon animation by faking the weapon's current clip in case the weapon has a different maxclip from its original, and is actually holding the gun with `clip == newmaxclip`.

You can see the conditions in the files changed.

I might say it needs tests 👍

### DIFF
--- a/regamedll/dlls/client.cpp
+++ b/regamedll/dlls/client.cpp
@@ -4824,7 +4824,7 @@ int EXT_FUNC GetWeaponData(edict_t *pEdict, struct weapon_data_s *info)
 					item->iuser1 = weapon->m_iSwing;
 
 #ifdef REGAMEDLL_FIXES
-					if (pPlayerItem == pPlayer->m_pActiveItem && weapon->m_iClip == II.iMaxClip)
+					if (pPlayerItem == pPlayer->m_pActiveItem && !weapon->m_fInReload && weapon->m_iClip == II.iMaxClip)
 					{
 						const WeaponInfoStruct *wpnInfo = GetDefaultWeaponInfo(II.iId);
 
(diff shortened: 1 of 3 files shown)


RauliTop:
Just for documentation, same fix from a metamod plugin: https://goldsrc.ru/threads/4165/

Also we have a video: https://www.youtube.com/watch?v=k_jGf7mLYBo
Bug happen when the weapon has 35 ammo and tries to reload.

Vaqtincha:
Спиздил мой метод и даже не указал на кредитах. 😀

dystopm:
> Спиздил мой метод и даже не указал на кредитах. 😀

Excuse me? I don't even know that forum 😆

I discovered this in a +10 year old plugin that describes this bug behaviour, which I was using in my weapon addons: https://forums.alliedmods.net/showthread.php?t=82093
```c++
	if( iButton & IN_RELOAD && !fInReload )
	{
		if( iClip >= iMaxClip )
		{
			set_pev(id, pev_button, iButton & ~IN_RELOAD)
			if( SILENT_BS & (1<<iId) && !get_pdata_int(iEnt, m_fSilent, XTRA_OFS_WEAPON) )
			{
				SendWeaponAnim( id, iId == CSW_USP ? 8 : 7 )
			}
			else
			{
				SendWeaponAnim(id, 0)
			}
		}
```
The only way to fake client weapon data is by GetWeaponData function, which is way better than resetting weaponanim. 

Sorry for not reading the entire internet to avoid a kid's cry. I suggest you to be polite, this isn't a forum, nor a plugin submission system.

PR #864, opened 2023-09-25
Refactored RemovePlayerItemEx and Extended DestroyItem in CBasePlayerItem

Javekson:
Extension of the DestroyItem method for the CBasePlayerItem class
- Added a boolean value to indicate the success of item deletion
- Implemented a condition for handling WEAPON_C4 specifically
- Added a condition to check for the presence of primary weapons

Updated the item removal mechanism in the RemovePlayerItemEx method
- Fixed a double hook invocation issue for grenades and performed minor code refactoring

Fix duplicate remove of item_thighpack


hajimura:
Nice catch!

PR #880, opened 2023-10-12
Update m_fLastMovement when player sent message

JUNESYNGOTOFLEX:


Vaqtincha:
saying is not movement 😄

JUNESYNGOTOFLEX:
> saying is not movement 😄 

if it is used for mp_autokick, shouldn't chatting mean that the player is in the game?

Vaqtincha:
@JUNESYNGOTOFLEX The (bad) player will launch a script or autoclicker and flood the chat (every n seconds) and the game will not kick him. Do you really need this?

JUNESYNGOTOFLEX:
@Vaqtincha I think you're right, it's still better to implement it through the 3rd party, on some mods it is necessary

Vaqtincha:
@JUNESYNGOTOFLEX you can check m_flLastTalk

justgo97:
> @JUNESYNGOTOFLEX The (bad) player will launch a script or autoclicker and flood the chat (every n seconds) and the game will not kick him. Do you really need this?

By the same logic a bad player can make script to trigger something to reset m_fLastMovement
I think editing this line https://github.com/s1lentq/ReGameDLL_CS/blob/b53ebb203794446b6ccb35d0ca975b6eef3de93d/regamedll/dlls/player.cpp#L4563 to take into account m_flLastTalk would be ideal, it would be annoying to get kicked while writing a long essay about whatever.

PR #886, opened 2023-10-28
New CVars: `mp_freezetime_duck` and `mp_freezetime_jump` 

FEDERICOMB96:
- Added cvar `mp_freezetime_duck` which allow ducking during freezetime.
`(0)` disabled
`(1)` enabled (default behaviour)

- Added cvar `mp_freezetime_jump` which allow jumping during freezetime. 
`(0)` disabled
`(1)` enabled (default behaviour)

Tested, work as intended 😄 

PS: Cvar's names/description corrections are welcome

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,8 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_dying_time                      | 3.0     | 0.0 | -            | Time for switch to free observing after death.<br/>`0` - disable spectating around death.<br/>`>0.00001` - time delay to start spectate.<br/>`NOTE`: The countdown starts when the player’s death animation is finished. |
 | mp_deathmsg_flags                  | 7       | 0   | 7            | Sets a bitsum for extra information in the player's death message.<br/>`0` disabled<br/>`1` position where the victim died<br/>`2` index of the assistant who helped the attacker kill the victim<br/>`4` rarity classification bits, e.g., `blinkill`, `noscope`, `penetrated`, etc. |
 | mp_assist_damage_threshold         | 40      | 0   | 100          | Sets the percentage of damage needed to score an assist. |
-| mp_prevent_duck_during_freezetime  | 0       | 0   | 1            | When set, players can't duck during freezetime.<br/> `0` disabled<br/>`1` enabled |
-| mp_prevent_jump_during_freezetime  | 0       | 0   | 1            | When set, players can't jump during freezetime.<br/> `0` disabled<br/>`1` enabled |
+| mp_freezetime_duck                 | 1       | 0   | 1            | When set, players can duck during freezetime.<br/> `0` disabled<br/>`1` enabled |
+| mp_freezetime_jump                 | 1       | 0   | 1            | When set, players can jump during freezetime.<br/> `0` disabled<br/>`1` enabled |
 </details>
 
 ## How to install zBot for CS 1.6?
--- a/README.md
+++ b/README.md
@@ -113,8 +113,8 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_dying_time                      | 3.0     | 0.0 | -            | Time for switch to free observing after death.<br/>`0` - disable spectating around death.<br/>`>0.00001` - time delay to start spectate.<br/>`NOTE`: The countdown starts when the player’s death animation is finished. |
 | mp_deathmsg_flags                  | 7       | 0   | 7            | Sets a bitsum for extra information in the player's death message.<br/>`0` disabled<br/>`1` position where the victim died<br/>`2` index of the assistant who helped the attacker kill the victim<br/>`4` rarity classification bits, e.g., `blinkill`, `noscope`, `penetrated`, etc. |
 | mp_assist_damage_threshold         | 40      | 0   | 100          | Sets the percentage of damage needed to score an assist. |
-| mp_freezetime_duck                 | 1       | 0   | 1            | When set, players can duck during freezetime.<br/> `0` disabled<br/>`1` enabled |
-| mp_freezetime_jump                 | 1       | 0   | 1            | When set, players can jump during freezetime.<br/> `0` disabled<br/>`1` enabled |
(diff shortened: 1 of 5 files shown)


abecee:
That's a great idea. Good job.

justgo97:
Instead of adding extra checks in PM_Duck & PM_Jump, It probably would be better to use PLAYER_PREVENT_DUCK 
 & PLAYER_PREVENT_JUMP flags on player when freezetime is on and to remove the flags once freezetime is over.

dystopm:
> Instead of adding extra checks in PM_Duck & PM_Jump, It probably would be better to use PLAYER_PREVENT_DUCK & PLAYER_PREVENT_JUMP flags on player when freezetime is on and to remove the flags once freezetime is over.

Those are meant to be using by modders. By enabling/disabling you may interfere with 3rd party plugin behaviours. (I suggested it)

dystopm:
The cvar name is too long and literal. It should have a short name instead, there isn't a reason to explain the control behaviour in the name. Something like `mp_freezetime_duck` and `mp_freezetime_jump`, you can see all other added cvars in game.cfg which are mostly compact based on what they control.

justgo97:
> > Instead of adding extra checks in PM_Duck & PM_Jump, It probably would be better to use PLAYER_PREVENT_DUCK & PLAYER_PREVENT_JUMP flags on player when freezetime is on and to remove the flags once freezetime is over.
> 
> Those are meant to be using by modders. By enabling/disabling you may interfere with 3rd party plugin behaviours. (I suggested it)

IMO Modders can enforce the cvars to be 0 or reset the flags on freezetime start/end so it won't really bother them much. It would be cleaner to not bloat PM_Duck & PM_Jump with these checks. Performance wise it's also better to avoid adding code in these functions that get called frequently. but it's fine to keep it the way it is now too

dystopm:
> > > Instead of adding extra checks in PM_Duck & PM_Jump, It probably would be better to use PLAYER_PREVENT_DUCK & PLAYER_PREVENT_JUMP flags on player when freezetime is on and to remove the flags once freezetime is over.
> > 
> > 
> > Those are meant to be using by modders. By enabling/disabling you may interfere with 3rd party plugin behaviours. (I suggested it)
> 
> IMO Modders can enforce the cvars to be 0 or reset the flags on freezetime start/end so it won't really bother them much. 

And yet, you're forcing modifications to handle this new behaviour which in the way it is now it wont hurt anybody. Instead, by using those custom flags, you can mess with more than one custom behaviour. That's why they are split

> It would be cleaner to not bloat PM_Duck & PM_Jump with these checks. 

Bloat why? They are two simple conditionals which are described in the comments, where one is a rule (cvar, like many others) and the other a custom flag. There are nastier conditions around the code, this one is completely sane

> Performance wise it's also better to avoid adding code in these functions that get called frequently. 

Hmm, performance card arguing on a simple statement? On C++? A machine-level interpreted source? 😛

FEDERICOMB96:
> The cvar name is too long and literal. It should have a short name instead, there isn't a reason to explain the control behaviour in the name. Something like `mp_freezetime_duck` and `mp_freezetime_jump`, you can see all other added cvars in game.cfg which are mostly compact based on what they control.

Good point. Ty ❤️

MatiasEsf:
> > The cvar name is too long and literal. It should have a short name instead, there isn't a reason to explain the control behaviour in the name. Something like `mp_freezetime_duck` and `mp_freezetime_jump`, you can see all other added cvars in game.cfg which are mostly compact based on what they control.
> 
> Good point. Ty ❤️

Screenshot_1

justgo97:
> And yet, you're forcing modifications to handle this new behaviour which in the way it is now it wont hurt anybody. Instead, by using those custom flags, you can mess with more than one custom behaviour. That's why they are split

Actually this current approach is less flexible from a modder POV, let's say I want to make only a specific team able to duck/jump during freeze time, if the flags were used I would just change the flags on player spawn with my desired value but since the flags are overrided by the cvar value there is nothing I can do other than setting the cvars to 0 and making my own implementation.

> Bloat why? They are two simple conditionals which are described in the comments, where one is a rule (cvar, like many others) and the other a custom flag. There are nastier conditions around the code, this one is completely sane
> 
> > Performance wise it's also better to avoid adding code in these functions that get called frequently.
> 
> Hmm, performance card arguing on a simple statement? On C++? A machine-level interpreted source? 😛

It's better to sperate concerns when possible, things like freezetime is only related to the first 10-5 seconds of the round, keeping them in a code that runs every frame for every player is not the optimal way to do this, this is just a suggestion, I'm not arguing lol 😛

dystopm:
> > And yet, you're forcing modifications to handle this new behaviour which in the way it is now it wont hurt anybody. Instead, by using those custom flags, you can mess with more than one custom behaviour. That's why they are split
> 
> Actually this current approach is less flexible from a modder POV, let's say I want to make only a specific team able to duck/jump during freeze time, if the flags were used I would just change the flags on player spawn with my desired value but since the flags are overrided by the cvar value there is nothing I can do other than setting the cvars to 0 and making my own implementation.
> 
> > Bloat why? They are two simple conditionals which are described in the comments, where one is a rule (cvar, like many others) and the other a custom flag. There are nastier conditions around the code, this one is completely sane
> > > Performance wise it's also better to avoid adding code in these functions that get called frequently.
> > 
> > 
> > Hmm, performance card arguing on a simple statement? On C++? A machine-level interpreted source? 😛
> 
> It's better to sperate concerns when possible, things like freezetime is only related to the first 10-5 seconds of the round, keeping them in a code that runs every frame for every player is not the optimal way to do this, this is just a suggestion, I'm not arguing lol 😛

> let's say I want to make only a specific team able to duck/jump during freeze time

Then don't use this cvar and code the rule? lmao, with or without this commit you'll be in need to do it

> keeping them in a code that runs every frame for every player is not the optimal way to do this

Seriously? Are we in 2005's alliedmods forums?

fred0r:
@FEDERICOMB96 
you changed the description of the cvar to the opposite, but not the function itself.
to currently deactivate jump/duck you now need to set its value to `1` - but your descriptions says:
```
// Allow players to duck during freezetime
// 0 - disabled

// Allow players to jump during freezetime
// 0 - disabled
```

Issue #887, opened 2023-10-29
`IsInWorld()` should use `sv_maxvelocity` cvar value instead of hardcoded number


Nord1cWarr1or:
https://github.com/s1lentq/ReGameDLL_CS/blob/96e2121c4504d26376d655f036d32e2fd967e4e5/regamedll/dlls/cbase.cpp#L869

Nord1cWarr1or:
https://github.com/s1lentq/ReGameDLL_CS/blob/96e2121c4504d26376d655f036d32e2fd967e4e5/regamedll/dlls/cbase.cpp#L869

StevenKal:
Yes, and will allow us (if we increase "sv_maxvelocity") to fix grenades removed if we throw them faster then 2000 units/sec (case with speedhack plugin or super fast LongJump tuned).

dystopm:
Seems reasonable

Nord1cWarr1or:
> Yes, and will allow us (if we increase "sv_maxvelocity") to fix grenades removed if we throw them faster then 2000 units/sec (case with speedhack plugin or super fast LongJump tuned).

It will not be a complete solution to the problem

PR #888, opened 2023-10-29
Fixed grenades disappearing when speed exceeds 2000 fixed units ignoring sv_maxvelocity

dystopm:
Issue #887 

Added cvar pointer caching of sv_maxvelocity. Did a little research on engine and haven't found more hardcoded usages of 2000 units.

### DIFF
--- a/regamedll/dlls/ggrenade.cpp
+++ b/regamedll/dlls/ggrenade.cpp
@@ -772,6 +772,13 @@ void CGrenade::BounceSound()
 
 void CGrenade::TumbleThink()
 {
+#ifdef REGAMEDLL_FIXES
+	if (pev->velocity.IsLengthGreaterThan(g_psv_maxvelocity->value))
+	{
+		pev->velocity = pev->velocity.Normalize() * g_psv_maxvelocity->value;
+	}
+#endif
+
 	if (!IsInWorld())
 	{
 		UTIL_Remove(this);
@@ -809,6 +816,13 @@ void CGrenade::TumbleThink()
 
 void CGrenade::SG_TumbleThink()
 {
+#ifdef REGAMEDLL_FIXES
+	if (pev->velocity.IsLengthGreaterThan(g_psv_maxvelocity->value))
+	{
+		pev->velocity = pev->velocity.Normalize() * g_psv_maxvelocity->value;
+	}
+#endif
+
 	if (!IsInWorld())
 	{
 		UTIL_Remove(this);
@@ -1322,6 +1336,13 @@ void AnnounceFlashInterval(float interval, float offset)
 
 void CGrenade::C4Think()
 {
+#ifdef REGAMEDLL_FIXES
+	if (pev->velocity.IsLengthGreaterThan(g_psv_maxvelocity->value))
+	{
+		pev->velocity = pev->velocity.Normalize() * g_psv_maxvelocity->value;
+	}
+#endif
+
 	if (!IsInWorld())
 	{
 #ifdef REGAMEDLL_FIXES
--- a/regamedll/dlls/cbase.cpp
+++ b/regamedll/dlls/cbase.cpp
@@ -865,12 +865,18 @@ BOOL CBaseEntity::IsInWorld()
 		return FALSE;
 	}
 
+#ifdef REGAMEDLL_FIXES
+	double maxvel = g_psv_maxvelocity->value;
+#else
+	double maxvel = 2000.0;
+#endif
+
 	// speed
-	if (pev->velocity.x >= 2000.0 || pev->velocity.y >= 2000.0 || pev->velocity.z >= 2000.0)
+	if (pev->velocity.x >= maxvel || pev->velocity.y >= maxvel || pev->velocity.z >= maxvel)
 	{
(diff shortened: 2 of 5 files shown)


justgo97:
You probably can use the cvar pointer here too:
[
](https://github.com/s1lentq/ReGameDLL_CS/blob/96e2121c4504d26376d655f036d32e2fd967e4e5/regamedll/dlls/gib.cpp#L8)

dystopm:
> You probably can use the cvar pointer here too: [line](https://github.com/s1lentq/ReGameDLL_CS/blob/96e2121c4504d26376d655f036d32e2fd967e4e5/regamedll/dlls/gib.cpp#L8)

Check Files changed, done in first place

justgo97:
> > You probably can use the cvar pointer here too: [line](https://github.com/s1lentq/ReGameDLL_CS/blob/96e2121c4504d26376d655f036d32e2fd967e4e5/regamedll/dlls/gib.cpp#L8)
> 
> Check Files changed, done in first place

oh Yeah didn't see it, well done

regamedll/dlls/ggrenade.cpp
```diff
@@ -772,6 +772,13 @@ void CGrenade::BounceSound()
 
 void CGrenade::TumbleThink()
 {
+#ifdef REGAMEDLL_FIXES
+	if (pev->velocity.IsLengthGreaterThan(g_psv_maxvelocity->value))
+	{
+		pev->velocity = pev->velocity.Normalize() * g_psv_maxvelocity->value;
```
justgo97, line 778: Correct me if I'm wrong but shouldn't pev->velocity be capped to the exact value of g_psv_maxvelocity->value while respecting the sign?

dystopm: Lengths and distances are positive numbers my friend. It's just math

dystopm: You can guess what happens when you normalize a vector with negative values, and then an scalar (another distance) is applied.

dystopm: First, I guess you wanted something like SV_CheckVelocity which clamps axles separately. Thing is, that function is out of contexts and it will apply the same solution to every entity with velocity out of bounds. The problem with this, is the fact grenade's direction are going to mess up in case one of their velocity axles goes above maxvelocity. My approach can clamp grenade's velocity without neccesarily having a velocity axis value above 2000 units. but will keep grenade direction smoothly and "natural" besides the fact it can't go at higher speed, and IsInWorld wont filter it.
Second, velocity's axles can be negative, but distances will be positive (- * - = + ?), and normalizing that vector and fixing its magnitude to the sv_maxvelocity value will keep the trajectory of the grenade on its current direction as I said before. Strictly clamping velocity axles to sv_maxvelocity is a "general" solution applied by the engine, but in this case, where we know the context, that potential fix could derive in a ugly way to force a grenade to stay in world.
By the way you can see gibs actually have this logic to keep their velocity vector without changing direction. I'm open to suggestions, but IMO this is a good and smooth approach to get rid of that limit.

justgo97: I think this new behavior will change the expected travel distance of the grenade and thus at some point direction will be different too, I suggest that making this new behavior under REGAMEDLL_ADD macro and adding a cvar to turn it on/off, And since you already added sv_maxvelocity usage to `IsInWorld` people can increase sv_maxvelocity to a high value like 9999 and then this won't be necessary and the problem of grenades disappearing will also be fixed, so maybe split this into two PRs, keeping this one for just fixing IsInWorld and another for new behavior you want to add with a cvar to turn it on/off.

Nord1cWarr1or: > people can increase sv_maxvelocity to a high value like 9999 and then this won't be necessary

1) Many people don't realize that grenades are missing due to over velocity. This should be fixed for everyone.
2) It is logical that any entity should be limited by cvar maxvelocity, grenades should not be an exception.

dystopm: > I think this new behavior will change the expected travel distance of the grenade and thus at some point direction will be different too, I suggest that making this new behavior under REGAMEDLL_ADD macro and adding a cvar to turn it on/off, 

I don't know, it seems you understand nothing from my explanation or you are not realizing how the math is applied. I also exposed this to @s1lentq to know his opinion and he didn't hesitate in figuring it is a correct approach, which is "the gib behaviour". **YOUR** approach of clamping vectors will mess up the directions. I invite you to test it yourself

> And since you already added sv_maxvelocity usage to `IsInWorld` people can increase sv_maxvelocity to a high value like 9999 and then this won't be necessary and the problem of grenades disappearing will also be fixed, 

By just incrementing it, wont mean it will be fixed because I'm just expanding the possibilities. Velocity can go higher numbers, and there's no reason to make everybody alter that cvar. This wont break non-modded scenarios, grenades will keep their trajectory, I'm just adding a rule to a boundary

> so maybe split this into two PRs, keeping this one for just fixing IsInWorld and another for new behavior you want to add with a cvar to turn it on/off.

I insist with my first paragraph

PR #893, opened 2023-11-11
Changed the order of setting pev->body for the correct value in SetBo…

Javekson:
…mbIcon()


dystopm:
Also check C4::PrimaryAttack on wpn_c4.cpp

Javekson:
> Also check C4::PrimaryAttack on wpn_c4.cpp

Could you provide more details on what's wrong with `C4::PrimaryAttack` in `wpn_c4.cpp`? It seems that `pev->body` is set there before `SetBombIcon`, which is the desired behavior, isn't it?

dystopm:
> > Also check C4::PrimaryAttack on wpn_c4.cpp
> 
> Could you provide more details on what's wrong with `C4::PrimaryAttack` in `wpn_c4.cpp`? It seems that `pev->body` is set there before `SetBombIcon`, which is the desired behavior, isn't it?

You're right! Mistake by watching my sources instead of this PR

PR #894, opened 2023-11-11
Implements Observer_Think Hook

Mythlogic:


dystopm:
Hook additions should be appended to the end of the virtual table, but everything else is alright

PR #895, opened 2023-11-13
feat(SwitchTeam): add of bomb removal on team switch

Javekson:


regamedll/dlls/player.cpp
```diff
@@ -8439,6 +8424,20 @@ void CBasePlayer::__API_HOOK(SwitchTeam)()
...
 	int NumDeadCT, NumDeadTerrorist, NumAliveTerrorist, NumAliveCT;
 	CSGameRules()->InitializePlayerCounts(NumAliveTerrorist, NumAliveCT, NumDeadTerrorist, NumDeadCT);
+
+	if (m_bHasC4)
+	{
+		if (NumAliveTerrorist > 0 && CSPlayer()->RemovePlayerItemEx("weapon_c4", true)) {
+			m_bHasC4 = false;
+			pev->body = 0;
+			SetBombIcon(FALSE);
+			SetProgressBarTime(0);
```
justgo97, line 8434:
```
pPlayer->m_bHasC4 = false;
pPlayer->pev->body = 0;
pPlayer->SetBombIcon(FALSE);
pPlayer->SetProgressBarTime(0);
```

You probably can remove these 4 lines since `RemovePlayerItemEx` already has them.

Javekson: Thank you for the feedback! I'll remove the suggested 4 lines. Regarding bRemoveAmmo, I'll keep it as true because in certain non-standard scenarios or mods, there might be a possibility of having ammo reserves for the bomb. I've tested this case.

PR #898, opened 2023-11-26
Update studio.h constants

dystopm:
- Updated studio.h header using ReHLDS files:
    - Limits increased according to engine ones
    - Added missing structures (in case of)
- `g_pBoneTransform` size using magic number changed to a defined number

### DIFF
--- a/regamedll/dlls/animation.cpp
+++ b/regamedll/dlls/animation.cpp
@@ -12,7 +12,7 @@ server_studio_api_t IEngineStudio;
 studiohdr_t *g_pstudiohdr;
 
 float (*g_pRotationMatrix)[3][4];
-float (*g_pBoneTransform)[128][3][4];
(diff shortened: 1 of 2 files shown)


regamedll/dlls/animation.cpp
```diff
@@ -12,7 +12,7 @@ server_studio_api_t IEngineStudio;
 studiohdr_t *g_pstudiohdr;
 
 float (*g_pRotationMatrix)[3][4];
-float (*g_pBoneTransform)[128][3][4];
+float (*g_pBoneTransform)[STUDIO_NUM_HULLS][3][4];
```
s1lentq, line 15: It's not really for that, should be use `MAXSTUDIOBONES` macro

dystopm: done!

PR #901, opened 2023-11-28
Initialize m_pevLastInflictor to nullptr to avoid garbage memory

dystopm:
@Nord1cWarr1or

### DIFF
--- a/regamedll/public/regamedll/API/CSEntity.h
+++ b/regamedll/public/regamedll/API/CSEntity.h
@@ -33,7 +33,8 @@ class CCSEntity
 {
 public:
 	CCSEntity() :
-		m_pContainingEntity(nullptr)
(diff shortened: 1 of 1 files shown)


Nord1cWarr1or:
I will test it

dystopm:
> I will test it

Went good?

Nord1cWarr1or:
@dystopm It seems to be fixed, I don't have any more server crashes, at least not yet

dystopm:
> @dystopm It seems to be fixed, I don't have any more server crashes, at least not yet

Lets take some time. But I'm sure it is related to this.

Nord1cWarr1or:
Still no crash. I think this option really helped

dystopm:
> Still no crash. I think this option really helped

I think we're done with this, @s1lentq

PR #858, opened 2023-09-07
Add an extended player's DeathMsg message

s1lentq:
**Description**
These changes introduces enhancements to the DeathNotice function,
providing extra death flags to convey more information in death messages.
These flags are used to include extra details about player deaths, making death messages more informative

- Extra Death Flags:

  - PLAYERDEATH_POSITION (Bit 1):
    Position of Victim's Death: This flag includes the coordinates (X, Y, Z) of the place where the victim died.
    It is useful for displaying a 'dead icon' on the HUD radar, indicating the victim's death location.

  - PLAYERDEATH_ASSISTANT (Bit 2):
    Index of Assistant: When set, this flag includes the index of the assisting player who helped the attacker kill the victim. It allows players to know who assisted in the kill.

  - PLAYERDEATH_KILLRARITY (Bit 4):
    Kill Rarity Classification: This flag adds a rarity classification to the death message.
    It includes information about the rarity of the kill, such as headshot kills, kills through walls, flash assists, and more.
    These classifications make death messages more engaging by highlighting exceptional kills.

    - Headshot Kill: Indicates that the victim was killed by a headshot, signifying exceptional accuracy and precision.

    - No-Scope Kill: Denotes a kill with a sniper rifle without using the scope, showcasing remarkable skill and reflexes.

    - Penetrated Kill: Highlights kills made through walls or objects, demonstrating tactical prowess.

    - Flash Assist: Marks kills assisted by flashbang grenades, revealing coordinated teamwork.

    - And More: Additional classifications may be introduced to further enrich the gameplay experience.

**Usage for Developers:**
AMXX plugin developers can use these flags to extract valuable information, such as kill rarity and other details, for use in their plugins.
Functions like hookchain CSGameRules_SendDeathMessage, register_message and register_event can be employed to access this additional information. 

**Benefits client-side for unofficial clients:**
These enhancements extend beyond the official game, but it can benefit unofficial game clients as well. Non-official clients can use this additional information to create visual enhancements and customizations. This allows for a richer gaming experience and increased player engagement.

PR #904, opened 2023-12-04
Add member m_iGibDamageThreshold to control GIB damage threshold

dystopm:
The only way to "naturally" GIB on death for players is reaching -9000 health points. This PR adds a CCSPlayer member that controls that value individually per player. It defaults to -9000.

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -2455,7 +2455,7 @@ void EXT_FUNC CBasePlayer::__API_HOOK(Killed)(entvars_t *pevAttacker, int iGib)
 		HintMessage("#Hint_cannot_play_because_tk", TRUE, TRUE);
 	}
 
-	if ((pev->health < -9000 && iGib != GIB_NEVER) || iGib == GIB_ALWAYS)
+	if (ShouldGibPlayer(iGib))
 	{
 
 #ifndef REGAMEDLL_FIXES
--- a/regamedll/dlls/player.h
+++ b/regamedll/dlls/player.h
@@ -641,6 +641,7 @@ public:
 	bool ShouldToShowAccount(CBasePlayer *pReceiver) const;
 	bool ShouldToShowHealthInfo(CBasePlayer *pReceiver) const;
 	const char *GetKillerWeaponName(entvars_t *pevInflictor, entvars_t *pevKiller) const;
+	bool ShouldGibPlayer(int iGib);
 
 	CBasePlayerItem *GetItemByName(const char *itemName);
 	CBasePlayerItem *GetItemById(WeaponIdType weaponID);
--- a/regamedll/dlls/player.h
+++ b/regamedll/dlls/player.h
@@ -448,7 +448,7 @@ public:
 	edict_t *EntSelectSpawnPoint_OrigFunc();
 	void PlayerDeathThink_OrigFunc();
 	void Observer_Think_OrigFunc();
-	
+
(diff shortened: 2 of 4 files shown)


dystopm:
> The function "CBasePlayer::ShouldGibPlayer" should probably not be in "CSPlayer.h" (more on "player.h"), more due to the fact it is not only "API only feature", due to internal "REGAMEDLL_API" define.

Thing is, I don't know why it doesn't compile when I put this inline function after the CSPlayer method declaration inside player.h - by the way, @s1lentq implemented the `DmgPenetrationLevel` methods in CSEntity.h (#858) when their declarations are on cbase.h and I guess it's because the same compilaton (also did the same in (#896).

StevenKal:
Maybe putting it inside the "#ifdef REGAMEDLL_API" on "player.h" line 953 could skip the compiler error, then a #else with duplicated function without the member.
However, a non-inline function, but does not worth to expose such tiny check.

dystopm:
> Maybe putting it inside the "#ifdef REGAMEDLL_API" on "player.h" line 953 could skip the compiler error, then a #else with duplicated function without the member. However, a non-inline function, but does not worth to expose such tiny check function.

```c++
#ifdef REGAMEDLL_API
inline CCSPlayer *CBasePlayer::CSPlayer() const {
	return reinterpret_cast<CCSPlayer *>(this->m_pEntity);
}
#endif
inline bool CBasePlayer::ShouldGibPlayer(int iGib)
{
#ifdef REGAMEDLL_API
	int threshold = CSPlayer()->m_iGibDamageThreshold;
#else 
	int threshold = GIB_PLAYER_THRESHOLD;
#endif
	return ((pev->health < threshold && iGib != GIB_NEVER) || iGib == GIB_ALWAYS);
}
```
It wont compile 💀 am I forgetting something?

StevenKal:
I was thinked to more something like:
```
#ifdef REGAMEDLL_API
inline CCSPlayer *CBasePlayer::CSPlayer() const {
	return reinterpret_cast<CCSPlayer *>(this->m_pEntity);
}

inline bool CBasePlayer::ShouldGibPlayer(int iGib)
{
	return ((pev->health < CSPlayer()->m_iGibDamageThreshold && iGib != GIB_NEVER) || iGib == GIB_ALWAYS);
}
#else
inline bool CBasePlayer::ShouldGibPlayer(int iGib)
{
	return ((pev->health < GIB_PLAYER_THRESHOLD&& iGib != GIB_NEVER) || iGib == GIB_ALWAYS);
}
#endif
```

But might not compile too!

dystopm:
> I was thinking to more something like:
> 
> ```c++
> #ifdef REGAMEDLL_API
> inline CCSPlayer *CBasePlayer::CSPlayer() const {
> 	return reinterpret_cast<CCSPlayer *>(this->m_pEntity);
> }
> 
> inline bool CBasePlayer::ShouldGibPlayer(int iGib)
> {
> 	return ((pev->health < CSPlayer()->m_iGibDamageThreshold && iGib != GIB_NEVER) || iGib == GIB_ALWAYS);
> }
> #else
> inline bool CBasePlayer::ShouldGibPlayer(int iGib)
> {
> 	return ((pev->health < GIB_PLAYER_THRESHOLD&& iGib != GIB_NEVER) || iGib == GIB_ALWAYS);
> }
> #endif
> ```
> 
> But might not compile too!

It wont recognize CSPlayer method, and I don't know why, it's getting defined above (lol). Some cpp coder may elaborate on this, I'm still a rookie :)

PR #908, opened 2023-12-13
New ConVar: `mp_defuser_allocation`

wopox1337:
from: https://totalcsgo.com/commands/mpdefuserallocation

```cpp
// Give defuser on player spawn
// 0 - No free defusers (default behavior)
// 1 - Random players
// 2 - All players
//
// Default value: "0"
mp_defuser_allocation  "0"
```

### DIFF
--- a/regamedll/dlls/multiplay_gamerules.cpp
+++ b/regamedll/dlls/multiplay_gamerules.cpp
@@ -2063,6 +2063,11 @@ void EXT_FUNC CHalfLifeMultiplay::__API_HOOK(RestartRound)()
 		GiveC4();
 	}
 
+#ifdef REGAMEDLL_ADD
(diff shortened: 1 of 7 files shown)


regamedll/dlls/multiplay_gamerules.cpp
```diff
@@ -5376,3 +5385,54 @@ void EXT_FUNC CHalfLifeMultiplay::__API_HOOK(SendDeathMessage)(CBaseEntity *pKil
 
 	MESSAGE_END();
 }
+
+void CHalfLifeMultiplay::GiveDefuserToRandomPlayer()
+{
+	int iDefusersToGive = 2;
```
mcs, line 5391: Is that number customizable? If not, why? A ratio would have been quite nice.

PR #925, opened 2024-01-17
Added support ccache for speedup recompilation

GermanAizek:
@wopox1337, @s1lentq more information:
https://ccache.dev/
https://en.wikipedia.org/wiki/Ccache


Garey27:
Hello, Ccache its indeed good solution for large projects if you delete "build" directory of CMake, but RegameDLL is not that large, and if you using it for developing its not will speedup too-much because CMake already build only changed files. Or maybe you can show some real case that Ccache is usefull?

PR #935, opened 2024-01-25
Restart trigger_multiple

overl4y:


s1lentq:
@overl4y Can you provide an map to demonstrate that restarting trigger is necessary?
need to be careful to this PR, because it can break behavior a lot of maps

overl4y:
> @overl4y Can you provide an map to demonstrate that restarting trigger_multiple is necessary? need to be careful to this PR, because it can break behavior a lot of maps

It's widely used in zombie escape maps to activate an **ambient_generic**, **func_door**, **func_breakable**, **func_tracktrain** and a **trigger_hurt** at the end of the map. But when the round ends quickly, it breaks the entire trigger_multiple flow.

Some examples:

ze_base_staff_lg
ze_biohazard_lab_dp
ze_breeze_warz_x1
ze_classic_oilrig
ze_coffeelobby_traintobusan_v2
ze_darkprofessional_dp1
ze_dasty_esc4
ze_death_heights_b5
ze_egypt_xmas_dp
ze_freezy_algeria
ze_frozen_throne_b3
ze_FTX_escape2_ma_fix
ze_last_battle_b2
ze_pathway_dp
ze_reverse_dust2_warz1

PR #943, opened 2024-02-08
Fix `SendDeathMessage` kill rarity flags and transform to VFUNC

FEDERICOMB96:
This solves some problems detected during the testing of the hook "RG_CSGameRules_SendDeathMessage" and is totally linked to https://github.com/s1lentq/reapi/pull/305

### DIFF
--- a/regamedll/dlls/gamerules.h
+++ b/regamedll/dlls/gamerules.h
@@ -741,7 +741,7 @@ public:
 	VFUNC bool HasRoundTimeExpired();
 	VFUNC bool IsBombPlanted();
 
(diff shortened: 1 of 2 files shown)


Nord1cWarr1or:
> some problems

Can you describe these problems?

FEDERICOMB96:
> > some problems
> 
> Can you describe these problems?

Modify `iRarityOfKill` and `iDeathMessageFlags` vars inside `SendDeathMessage` function, amxx hook can't detect it.

Replace `pVictim->m_bHeadshotKilled` with `iRarityOfKill & KILLRARITY_HEADSHOT` is only for complete functionality of native `rg_send_death_message` and being able to change its value through the hook.

PR #949, opened 2024-02-22
Implement `game_round_end` and `game_round_freeze_end` triggers

overl4y:


mcs:
Hi, I'm sadly quite new to the party. Does this change mean that plugins now can directly register on both events instead of looking for log message and/or calculate wait times based on freezetime?

PR #958, opened 2024-04-09
don't send radio message to teammate (if freeforall 1)

Vaqtincha:


Vaqtincha:
without fix zero length
```
public CBasePlayer_Radio(const pPlayer, const msg_id[], const msg_verbose[], pitch, bool:showIcon)
{
  if (!equal(msg_verbose, "#Fire_in_the_hole"))
    return HC_CONTINUE

  SetHookChainArg(3, ATYPE_STRING, "")

  return HC_CONTINUE
}

```

image

PR #963, opened 2024-05-25
Implement Check* Hooks

JohanCorn:
- CheckGameOver
- CheckTimeLimit
- CheckFragLimit
- CheckMaxRounds
- CheckWinLimit

### DIFF
--- a/regamedll/dlls/API/CAPI_Impl.cpp
+++ b/regamedll/dlls/API/CAPI_Impl.cpp
@@ -337,6 +337,12 @@ GAMEHOOK_REGISTRY(CBasePlayer_PlayerDeathThink);
 GAMEHOOK_REGISTRY(CBasePlayer_Observer_Think);
 GAMEHOOK_REGISTRY(CBasePlayer_RemoveAllItems);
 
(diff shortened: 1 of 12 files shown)


dystopm:
Question, why are you declaring hook function as VFUNC?

JohanCorn:
> Question, why are you declaring hook function as VFUNC?

I made a mistake, but thank you for pointing it out.

JohanCorn:
> Question, why are you declaring hook function as VFUNC?

I have changed the VFUNC BOOL declarations to BOOL. However, I'm not sure if it is also necessary to change the bool type BOOL (int) as I did. Could you please review this aspect as well? Thank you!

dystopm:
> > Question, why are you declaring hook function as VFUNC?
> 
> I have changed the VFUNC BOOL declarations to BOOL. However, I'm not sure if it is also necessary to change the bool type BOOL (int) as I did. Could you please review this aspect as well? Thank you!

Sorry for the late answer. Don't use BOOL, because it is the same as int, and they will not matching origin structure size (int=4, bool=1), leading to unexpected behaviour with addons. Original functions return type were bool -> keep bool :)

Issue #992, opened 2024-08-11
Sometimes the reload animation does not play


q52888940:
OMG, I always thought REGAMEDLL fixed it, I don't think we need to fix this with AMXX, maybe it's time to consider fixing it?
Sometimes reloading doesn't play the reloading action

dystopm:
On the one hand I understand what you are saying, but on the other hand you are not leaving any clues to be able to track the problem. It's kind of mandatory to give context IMO.

Versions of everything, any 3rd party plugin that may interfer, etc etc

q52888940:
> On the one hand I understand what you are saying, but on the other hand you are not leaving any clues to be able to track the problem. It's kind of mandatory to give context IMO.
> 
> Versions of everything, any 3rd party plugin that may interfer, etc etc

This plugin fixes it, I have no way to track down the cause of the problem, but this problem has been in this game for a long time, from the initial version to now, and it has not been fixed, sorry for my poor English, but I believe you can understand what I mean, in some cases the action sequence is not played when reloading weapons

q52888940:
[reload_animation_fix.txt](https://github.com/user-attachments/files/16576039/reload_animation_fix.txt)

dystopm:
I'll be glad to hear more opinions because I've never experienced (without ping issues) that reload animation is not playing. I mostly work with weapons when related to modding, so I might say I've got experience with it.

By looking at the code:

- I don't know why m_fInSpecialReload is not getting reset on weapon changing, I'll fix that
- Deagle is getting a fix on #985 
- I need to go deep this issue, because what plugin does is only reset weaponanim value to weapon's respective idle animation "after" reload, which is something that mod does normally. DefaultReload assigns a timer to m_flTimeWeaponIdle which later ensures weapon get idle animation - but this can not always work, because player can spam +attack leading to the weaponanim entvar to not get reset to 0 because of firing, and I'm not getting why that provokes (guessing by your thoughts) that reload animation is not playing correctly, which IS in fact. I need an opinion from client-side @s1lentq

Vaqtincha:
This bug has existed for a long time. If you shoot continuously (up to 0), the reloading animation sometimes does not play) And I think this is a client bug (since with cl_lw 0 this bug is not observed)

PR #993, opened 2024-08-11
Fix Shotguns reload flag not getting reset on weapon changing

dystopm:
-title-
Related to #992 
References: https://forums.alliedmods.net/showthread.php?p=1620401


q52888940:
all weapon? or just Shotguns

dystopm:
> all weapon? or just Shotguns

Only shotguns (m3, xm1014) handle their reload workflow with this specific member, check weapons.cpp and all related code inside

PR #919, opened 2024-01-07
Fix excessive punchangle when getting shield shot

dystopm:
Fixes #918 - it demonstrates how abrupt punchangle change was done when getting shot using tactical shield.

PR #999, opened 2024-08-16
Refactor Shield hit handling to remove redundancy and fix related bugs

dystopm:
## Purpose
The shield hit logic was duplicated in two separate functions: FireBullets3 (rifle/pistol bullets) and TraceAttack (specific case for shield hits from shotguns, or coming from FireBullets). Both contained the same bugs, and in a previous PR (#919) one of these bugs was fixed only in FireBullets3, leaving TraceAttack still affected.

Additionally, TraceAttack had a specific bug where the damage variable was being reset to 0 before updating the punch angles. This caused the X and Z punch angles to always be 0, regardless of damage, effectively nullifying any prior punch angle adjustments.

There was also an inconsistency in the clamping logic:
- For the X axis, only the lower bound (-4) was enforced, leaving the upper bound (4) unchecked.
- For the Z axis, both lower (-5) and upper (5) bounds were correctly enforced.

As a result, X axis values could exceed the intended range without restriction. This behavior was identified and corrected.

Another issue was that if a player’s punch angles exceeded the hardcoded thresholds (±4 for X and ±5 for Z), the values would be clamped directly, overwriting any previously assigned punch angles. This needed a safeguard to check whether the current punch angles were within the modifiable range before applying changes, to avoid unintentionally overwriting existing values.

## Approach
- Extracted and unified the shield hit handling logic from FireBullets3 and TraceAttack into a new HitShield function, extending the previous PR’s bug fixes (#919) so they now apply to both cases.

- Implemented a threshold check before modifying punch angles to ensure changes are only applied if the current values are within the range of -4/-5 to 4/5. This prevents overwriting valid existing punch angles outside the intended range.

- Fixed the missing upper bound check for the X axis by adding the 4 limit, ensuring it is clamped consistently with the lower bound.

- In TraceAttack, adjusted the order of operations so the damage variable is now reset after the HitShield logic, ensuring punch angle calculations use the correct damage value.


regamedll/dlls/cbase.cpp
```diff
@@ -1503,6 +1480,35 @@ VectorRef CBaseEntity::__API_HOOK(FireBullets3)(VectorRef vecSrc, VectorRef vecD
...
+		EMIT_SOUND(ENT(pev), CHAN_VOICE, "weapons/ric_metal-1.wav", VOL_NORM, ATTN_NORM);
+	else
+		EMIT_SOUND(ENT(pev), CHAN_VOICE, "weapons/ric_metal-2.wav", VOL_NORM, ATTN_NORM);
+
+	UTIL_Sparks(ptr->vecEndPos);
+
+#ifdef REGAMEDLL_FIXES
+	if (Q_fabs(pev->punchangle.x) < 4.0)
+	{
+		pev->punchangle.x = clamp(flDamage * RANDOM_FLOAT(-0.15, 0.15), -4.0f, 4.0f);
```
fl0werD, line 1495: I think it's useless check because clamp

regamedll/dlls/cbase.cpp
```diff
@@ -1503,6 +1480,35 @@ VectorRef CBaseEntity::__API_HOOK(FireBullets3)(VectorRef vecSrc, VectorRef vecD
...
+	}
+#else
+	pev->punchangle.x = flDamage * RANDOM_FLOAT(-0.15, 0.15);
+
+	if (pev->punchangle.x < 4)
+		pev->punchangle.x = -4;
+#endif
+
+#ifdef REGAMEDLL_FIXES
+	if (Q_fabs(pev->punchangle.z) < 5.0)
```
fl0werD, line 1505: I think it's useless check because clamp

dystopm:
Added more details to the PR description for clarity — feel free to check it out. @s1lentq @fl0werD

s1lentq (approved):

PR #1005, opened 2024-08-27
Add cvar for stamina restoration speed based on fps reference.

Garey27:
This PR fixes time to restore stamina after jump based on reference fps.

Which described on another PR:
https://github.com/s1lentq/ReGameDLL_CS/pull/755

After setting cvar mp_stamina_restore_rate to “100.0”, players with no matter how high (1000) or low (30) fps will have the same stamina recovery rate after a jump as they would have with 100 fps.


Vaqtincha:
ifdef REGAMEDLL_ADD forget ?

AlexSwartz:
Hello, I test this fix and there is some glitch when client FPS is greater than ~145FPS. When landing, there is a slight judder to the camera.

I tested with mp_stamina_restore_rate 100, with this merged to the latest release 5.26.0.668.

https://streamable.com/azz0gy

Garey27:
@AlexSwartz This because client's predict your movement with original PM_Move code.

PR #1008, opened 2024-08-30
API: Added new API funcs

lozatto:


dystopm:
I think most of them are a great addition and will help more than one coder, BTW I got some observations:

- There are functions here that are already exported. I must remind you that the ReGame API is not the only interface that is exported here: `enginefuncs_t` includes `SetSize`, `SetOrigin`, `PointContents` and `FindEntityInSphere`. I don't see the point in adding them redundantly.
- `UTIL_CountEntities` is a miscellaneous one (did myself) and can be done in any metamod/amxx subplugin even without using regame. There's a **NEED** in exporting it?
- Are you aware that `UTIL_IsSpawnPointOccupied` is bugged and is also not being used? 👀 (it needs NOXREF tag also) (VEC_SPOT_HULL_MIN/MAX vectors are not considering negative height)

PR #1009, opened 2024-09-02
Fixes wrong sending of the weaponmodel body group

ShadowsAdi:


dystopm:
Can you add a description of why is actually sending a wrong value? Just curious, I always knew that weapons can change their body values, some edited clients allows hands changing (like in the case of CSO weapon models)

fl0werD:
It's view model body. If I was on T team with C4 and custom view model with 2 bodies, then I will be hold custom view model with second body with this PR.

And also it broke prediction.

dystopm:
> It's view model body. If I was on T team with C4 and custom view model with 2 bodies, then I will be hold custom view model with second body with this PR.
> 
> And also it broke prediction.

That's why I asked that also, because player body differs from weapon body value. There's something unknown in this modification.

Vaqtincha:
scale

ShadowsAdi:
So you're saying that setting the animation based on the weapon's body is better, the expected behaviour, than sending the player's body group without breaking the client prediction?

PR #1012, opened 2024-09-15
Botescape

Vaqtincha:
Little support escape mode for bots...


Vaqtincha:
I haven't played this mode myself. but they tell me that it would be nice if the bots first looked for weapons on the map.

SergeyShorokhov:
@Vaqtincha In any case, we appreciate the addition of this mode for bots. After these changes we can at least try the mode (test) and add new functionality as we test it.

PR #1035, opened 2024-12-15
add new  cvar "bot_excellent_morale"

Vaqtincha:
Finally the bots don't sit around after a couple of losses :D


q52888940:
也许用控制台指令会更好，这打破了原有的规则

PR #1039, opened 2024-12-27
feat: add new cvar "mp_logkills"

golukon:
## Purpose
Add new cvar "mp_logkills". Kills are logging when set to 1, otherwise when set to 0


dystopm:
Use REGAMEDLL_ADD macro

golukon (commented):
@dystopm Ready

PR #1040, opened 2024-12-27
fix: calculate UpdateLocation fun not so often

golukon:
Fix mistake in UpdateLocation (its calculated a lot due to wrong condition)


dystopm:
Use REGAMEDLL_FIXES macro

golukon (commented):
@dystopm Ready

PR #1054, opened 2025-02-17
Bots won't react to player shooting at them while mp_friendlyfire is 0

Huga22118:
The bots will not react if they get shot by a teammate while mp_friendlyfire is 0.


hajimura:
Nice catch!

PR #1055, opened 2025-02-18
Bot Chatter: Bots will react to enemy snipers presence

Huga22118:
previously unused chatter on both 1.6 and czero but is used on cs source


sweetcarolinahoney:
Good stuff. Hope to see more useful things from CSS/CSGO, like avoiding enemy grenades and better grenade usage etc., since default bots' grenade behavior is very shitty.

PR #1066, opened 2025-04-06
`FIX`: Remove grenades and C4

GLoOoccK:


Vaqtincha:
So basically you haven't fixed anything.

PR #1069, opened 2025-04-13
API: Knockback

GLoOoccK:


dystopm:
Good improvement

regamedll/dlls/player.cpp
```diff
@@ -1246,8 +1246,16 @@ BOOL EXT_FUNC CBasePlayer::__API_HOOK(TakeDamage)(entvars_t *pevInflictor, entva
...
+				{
+					CBaseEntity *temp = CBaseEntity::Instance(pevAttacker);
+					Knockback(temp, knockback.value);
+				}
+#else
 				Vector attack_velocity = (pev->origin - pAttack->pev->origin).Normalize() * 170;
 				pev->velocity = pev->velocity + attack_velocity;
+#endif
 
 				m_flVelocityModifier = 0.65f;
```
justgo97, line 1260: Maybe include this line in the `Knockback` function since it reduces player speed

dystopm:
You can add a bool parameter to pass the precondition of the factors (from a large flinch or not) 😁

JUNESYNGOTOFLEX:
will it be able to extend to this part of the code?
https://github.com/rehlds/ReGameDLL_CS/blob/master/regamedll/cbase.cpp#L660

RauliTop:
Does anyone know why is using a check to velocity when is lower than 300?

https://github.com/GLoOoccK/ReGameDLL_CS/blob/f7891cb51b0233887e5a8dbf75788e13edce73b8/regamedll/dlls/player.cpp#L1247

@GLoOoccK it will be fine another cvar to change that value. What do you think?

PR #1083, opened 2025-06-17
Expose CheckWinConditions in API

jonatan1024:
## Purpose
I needed to manually trigger CheckWinConditions because I have players from other games in the server and they don't call CheckWinConditions on their death/disconnect, therefore the round wasn't ending properly.

## Approach
Added CheckWinCondition into the API and incremented the minor number.


jonatan1024:
Wow yes you are right, I didn't realize that the API provides GetGameRules. That is sufficient for my purposes. Closed by deleting the branch.

PR #1103, opened 2025-08-10
Prevent AddMultiDamage execution with zero/negative damage

dystopm:
## Purpose
The `AddMultiDamage` function can be skipped entirely when no damage will be applied. While some existing cases already avoid calling it with zero damage (e.g. `FireBullets3`, shield hit avoids `TraceAttack`->`AddMultiDamage` calls), there are other scenarios where it is still invoked with a `damage` value of `0` or less (e.g., hitting a player with shotgun pellets while they have an active Shield).

These unnecessary calls not only waste processing but can also have unintended side effects for APIs implemented by developers that rely on `TakeDamage` being called only when damage is actually applied. This PR ensures that the function only executes when `flDamage > 0`.

## Approach
Added a simple check at the start of `AddMultiDamage` to immediately return if the `flDamage` parameter is `0` or less.

### DIFF
--- a/regamedll/dlls/weapons.cpp
+++ b/regamedll/dlls/weapons.cpp
@@ -107,6 +107,11 @@ void EXT_FUNC __API_HOOK(AddMultiDamage)(entvars_t *pevInflictor, CBaseEntity *p
 	if (!pEntity)
 		return;
 
+#ifdef REGAMEDLL_FIXES
+	if (flDamage <= 0.0f) // avoid zero or negative damage TakeDamage
+		return;
+#endif
+
 	gMultiDamage.type |= bitsDamageType;
 
 	if (pEntity != gMultiDamage.hEntity)


regamedll/dlls/weapons.cpp
```diff
@@ -107,6 +107,11 @@ void EXT_FUNC __API_HOOK(AddMultiDamage)(entvars_t *pevInflictor, CBaseEntity *p
 	if (!pEntity)
 		return;
 
+#ifdef REGAMEDLL_FIXES
+	if (flDamage <= 0.0f) // avoid zero or negative damage TakeDamage
```
s1lentq, line 111: This fix makes sense.  
However, there is one detail I remember, some custom maps intentionally use negative damage values (e.g., `env_laser` with damage = -300) as a way to heal players.

The maps I recall include `bhop_zerocheaters`, `deathrun_forest`, `deathrun_friends`, but there are many more with different scenarios.

This is not a bug but rather a hidden feature that level designers sometimes rely on. With your change, such maps would stop working as intended, since TakeDamage would never be triggered.

So the fix could unintentionally break existing community content.  
One possible approach could be to allow negative values (skip only flDamage == 0).

PR #1106, opened 2025-08-14
Optimize TraceBleed to skip unnecessary TraceLine calls

dystopm:
## Purpose
Optimize `TraceBleed`, which is called when a player is shot, by reducing unnecessary `TraceLine` calls for blood decal placement. In high-fire-rate modded servers (e.g., zombie mods), this can otherwise consume unnecessary resources.

## Approach
Move the 1/3 probability check to the start of the loop in `TraceBleed` and continue early when it fails, avoiding the cost of computing jitter and calling `TraceLine` in those cases.


hajimura:
Nice catch! 👍

PR #1108, opened 2025-08-31
Enable shield HUD pickup icon via WeaponList/WeapPickup

dystopm:
## Purpose
Enable the unused shield HUD icon in CS 1.6. Unlike weapons, the shield is not normally sent through the `WeaponList` message, so its history icon never appears on pickup. However, the file `weapon_shieldgun.txt` exists in the game resources with HUD icon data that was never utilized.

## Approach
Extend `GiveShield` to send a `WeaponList` placeholder entry for the shield and trigger the `WeapPickup` message when the player receives it. This activates the existing HUD assets so the shield displays its history icon consistently, matching the behavior of other items.

## Learning
https://github.com/Velaron/cs16-client/blob/87518233ad5f35a9c0f6ea8919ba98c4145834b6/cl_dll/ammo.cpp#L631
https://github.com/Velaron/cs16-client/blob/main/cl_dll/ammohistory.h
https://github.com/Velaron/cs16-client/blob/87518233ad5f35a9c0f6ea8919ba98c4145834b6/cl_dll/ammo.cpp#L530

[](https://www.youtube.com/watch?v=i2GEziOBA_o)

### DIFF
--- a/regamedll/dlls/player.cpp
+++ b/regamedll/dlls/player.cpp
@@ -3412,7 +3412,23 @@ void EXT_FUNC CBasePlayer::__API_HOOK(GiveShield)(bool bDeploy)
 		}
 	}
 
-#ifndef REGAMEDLL_FIXES
+#ifdef REGAMEDLL_FIXES
+	MESSAGE_BEGIN(MSG_ONE, gmsgWeaponList, nullptr, pev);
+		WRITE_STRING("weapon_shieldgun");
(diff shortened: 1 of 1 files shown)


fl0werD:
I think you need to write WeaponList in that place with MSG_INIT.
https://github.com/rehlds/ReGameDLL_CS/blob/master/regamedll/dlls/client.cpp#L235

PR #1151, opened 2026-02-27
Allow reload on weapons with altered ammo type

BoloVEVO:
## Purpose
ReGameDLL already gives support via ReAPI to implement custom ammo types, making possible to change the ammo type of any weapon,  the current checks of weapon reload is that the player needs to have the original ammo type of the weapon in order to reload, which lead into issues if the user wishes to alter the current ammo type.

## Approach
It now checks with the current item's ammo type instead of a hardcoded ammo type.

If there's a particular reason why to check the reload function with a hardcoded ammo type instead of grabbing the current item's ammo type, please explain.


BoloVEVO:
I will reopen this pr again I forgot I did the pr with the master branch instead of a separate branch.

PR #1157, opened 2026-04-16
Allow reload on weapons with altered ammo type

BoloVEVO:
## Purpose
ReGameDLL already gives support via ReAPI to implement custom ammo types, making possible to change the ammo type of any weapon, the current checks of weapon reload is that the player needs to have the original ammo type of the weapon in order to reload, which lead into issues if the user wishes to alter the current ammo type.

## Approach
It now checks with the current item's ammo type instead of a hardcoded ammo type.

If there's a particular reason why to check the reload function with a hardcoded ammo type instead of grabbing the current item's ammo type, please explain.

### DIFF
--- a/regamedll/dlls/wpn_shared/wpn_ak47.cpp
+++ b/regamedll/dlls/wpn_shared/wpn_ak47.cpp
@@ -175,7 +175,7 @@ void CAK47::Reload()
 {
 #ifdef REGAMEDLL_FIXES
 	// to prevent reload if not enough ammo
-	if (m_pPlayer->ammo_762nato <= 0)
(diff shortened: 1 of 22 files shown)


dystopm:
First, there is no need to replace one member check with another. Those checks can simply be skipped, since the ammo-related validation is redundant and already performed inside DefaultReload.

More importantly, please avoid modifying the original disassembled code directly. The purpose of this repository structure is to preserve the original code as-is, and place fixes or behavioral changes under the REGAMEDLL_FIXES macro.

BoloVEVO:
Done, no issues so far.

PR #1163, opened 2026-07-03
Add `mp_knife_wall_sparks` cvar: spark effect when a knife hits a wall

belomaxorka:
## What

New cvar `mp_knife_wall_sparks` (default `0` — original behavior). When enabled, a spark effect (`TE_SPARKS`) is shown at the point of impact whenever a knife hits a wall — both for the slash (`CKnife::Swing`) and the stab (`CKnife::Stab`) attack, in the exact code branches where `weapons/knife_hitwall1.wav` is played.

## Why

This is a popular cosmetic tweak that servers currently implement with AMXX plugins that hook every `EmitSound` call and reconstruct the impact point indirectly (via `get_user_aiming` and distance heuristics), which is both inaccurate and wasteful. Native support is a one-liner per attack path using the existing `UTIL_Sparks()` utility:

- the exact impact point is taken from the attack `TraceResult`;
- the effect is sent as `MSG_PVS` (only to nearby players), not `MSG_ALL`;
- zero overhead when the cvar is disabled.

## Implementation notes

- Gated by `#ifdef REGAMEDLL_ADD`, vanilla builds are unaffected.
- Default `0` keeps the original behavior bit-for-bit.
- Documented in `README.md` and `dist/game.cfg`.

## Tested

Verified on a live ReHLDS server (Windows, CS 1.6): with `mp_knife_wall_sparks 1` sparks appear at the exact impact point for both attack buttons; with `0` behavior is identical to the original (sound only). The cvar can be toggled at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_playerid_showhealth             | 1       | 0   | 2            | Player ID display mode.<br/>`0` don't show health<br/>`1` show health for teammates only (default CS behaviour)<br/>`2` show health for all players |
 | mp_playerid_field                  | 3       | 0   | 3            | Player ID field display mode.<br/>`0` don't show additional information<br/>`1` show team name<br/>`2` show health percentage<br/>`3` show both team name and health percentage |
 | mp_knockback                       | 170     | -   | -            | Knockback force applied to the victim when damaged by strong weapons (e.g. `AWP`, `AK47`).<br/>Works only if not crouching, and not hit in the legs.<br/>Set to `0` to disable. |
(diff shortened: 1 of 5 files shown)


JUNESYNGOTOFLEX:
yes it is. when implementing this feature you may some difficulties when writing the AMXX plugin.
however it is important to consider the customization of certain servers where there is a need to separate specific players using this feature (for ex: in a zombie plague mod, where zombies (m_iTeam == TERRORIST) are not allowed to have this feature.
To support specific game features you can see how new members are implementing them (such as m_bPlantC4Anywhere to mp_plant_c4_anywhere, m_iWeaponInfiniteAmmo to mp_infinite_ammo, etc.)

I was thinking about the existing m_trHit which passes the tracehandle which is used to create the decal. if this is changed in wpn_knife.cpp it could be used to pass data such as
tr. pHit, vecEndPos, vecPlaneNormal.. to create any custom message effects and custom server modification checks.

PR #1164, opened 2026-07-03
Add `mp_show_bomb_timer` cvar: show C4 countdown on the HUD round timer

belomaxorka:
## What

New cvar `mp_show_bomb_timer` (default `0` - original behavior). When enabled, after the C4 is planted the HUD round timer shows the countdown to the explosion (driven by `m_flC4Blow`, the same value the game uses to detonate) instead of disappearing. After a successful defuse the timer switches back to the round time.

## How it works

The stock CS 1.6 client hides the HUD round timer when it receives `BombDrop` with the "planted" flag. Two things make the countdown display possible:

- `CBasePlayer::SyncRoundTimer()` now substitutes the remaining C4 time and precedes the `RoundTime` message with an empty `ShowTimer` message, which un-hides the timer on the client. Since `SyncRoundTimer` is the single place that drives the HUD timer, every existing sync point picks the feature up automatically: initial HUD sync on connect (`UpdateClientData`), respawn (`Spawn`), freeze end (`OnRoundFreezeEnd`).
- On plant, all players are resynced right **after** `gmsgBombDrop` is broadcast (`wpn_c4.cpp`), so within one frame the client processes: hide (BombDrop) ? show (ShowTimer) ? countdown value (RoundTime). No periodic resending is needed - the client counts down locally from the exact value.

Unlike the popular AMXX implementations (which hook the plant and re-send `ShowTimer` + `RoundTime` to all players every second forever), this is fully event-driven: two messages per player at plant/defuse and nothing in between.

## Edge cases covered

- **Same-frame defuse**: the just-defused C4 is marked `FL_KILLME` by `UTIL_Remove` before the resync, so the bomb lookup can never pick it up.
- **Aborted defuse** (released `+use` / defuser died): no resync, the countdown correctly keeps running.
- **Late join / reconnect / respawn while the bomb is ticking**: the initial HUD sync sends `ShowTimer` + the exact remaining time.
- **Round restart**: `CleanUpMap()` removes the C4 before any sync, plus the freeze-period guard - no stale bomb can leak into the new round.
- **`mp_roundtime 0` servers** (timer normally hidden as useless): the existing `HIDEHUD_TIMER` logic shows the timer for the countdown and hides it back after the defuse.
- **Explosion**: the client display reaches 0:00 exactly at the detonation, since both are driven by `m_flC4Blow`.
- The displayed value uses the same floor semantics as the stock round timer (the digit means "full seconds remaining").

## Behavior notes

- After a successful defuse the round timer is **restored**, not hidden. This is deliberate: with `mp_round_infinite` (flag `e`) or plugins that supersede the round end, the round continues after a defuse and the round time is meaningful again. On regular servers the round ends a few seconds later anyway and the next round resyncs everything - same visuals as any other round ending.
- If a plugin changes `m_flC4Blow` after the plant (ReAPI), the client countdown will not follow automatically - the plugin can trigger a resync itself; a periodic resync was intentionally avoided to keep the feature zero-cost.

## Tested

Verified on a live ReHLDS server (Windows, CS 1.6): plant (as T and observing as CT), successful defuse, aborted defuse, explosion, reconnect mid-countdown, `mp_show_bomb_timer 0` - behavior identical to the original.

---

**Edit:** restore the HUD round timer after the bomb explodes. If the bomb round-end check is blocked (`mp_round_infinite` flag `e`), the explosion does not end the round and the timer stayed frozen at 0:00; it is now resynced for all players in `CGrenade::Explode2`, mirroring the resync already done on defuse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_hostage_hurtable                | 1       | 0   | 1            | The hostages can take damage.<br/>`0` disabled<br/>`1` from any team<br/>`2` only from `CT`<br/>`3` only from `T` |
 | mp_show_radioicon                  | 1       | 0   | 1            | Show radio icon.<br/>`0` disabled<br/>`1` enabled |
 | mp_show_scenarioicon               | 0       | 0   | 1            | Show scenario icon in HUD such as count of alive hostages or ticking bomb.<br/>`0` disabled<br/>`1` enabled |
+| mp_show_bomb_timer                 | 0       | 0   | 1            | Show the time until the planted bomb explodes on the HUD round timer.<br/>`0` disabled<br/>`1` enabled (when the bomb is planted, the round timer shows the C4 countdown) |
 | mp_old_bomb_defused_sound          | 1       | 0   | 1            | Play "Bomb has been defused" sound instead of "Counter-Terrorists win" when bomb was defused<br/>`0` disabled<br/>`1` enabled |
 | showtriggers                       | 0       | 0   | 1            | Debug cvar shows triggers. |
 | sv_alltalk                         | 0       | 0   | 5            | When players can hear each other ([further explanation](../../wiki/sv_alltalk)).<br/>`0` dead don't hear alive<br/>`1` no restrictions<br/>`2` teammates hear each other<br/>`3` Same as 2, but spectators hear everybody<br/>`4` alive hear alive, dead hear dead and alive.<br/>`5` alive hear alive teammates, dead hear dead and alive.
--- a/dist/game.cfg
+++ b/dist/game.cfg
@@ -191,6 +191,13 @@ mp_show_radioicon "1"
 // Default value: "0"
 mp_show_scenarioicon "0"
 
+// Show the time until the planted bomb explodes on the HUD round timer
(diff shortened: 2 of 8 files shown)


sankalpg777:
Hello,

I've been testing the new `mp_show_bomb_timer` feature from #1164 on a regular CS 1.6 server.

The C4 countdown itself works very well, but I noticed that after the round result has already been decided, the HUD timer can remain visible during `mp_round_restart_delay`.

I think the expected behavior on a normal server should be:

1. **C4 explodes and the round ends**
   Hide the HUD timer until the next round starts.

2. **C4 is defused and the round ends**
   Hide the HUD timer until the next round starts.

In other words, once the round has actually entered its round-end/restart-delay state, the timer should disappear instead of showing the restored round timer.

I saw the note in #1164 explaining that restoring the round timer after defuse/explosion is intentional because `mp_round_infinite` or plugins may prevent the round from ending.

I agree with that behavior when the round actually continues.

My suggestion would therefore be:

* **If the round continues:** restore/show the appropriate timer.
* **If the round has ended:** hide the HUD timer until the next round.

This would also make the behavior consistent regardless of whether the round ended by bomb explosion or bomb defuse.

Tested with:

`mp_show_bomb_timer "1"`

Thanks for adding this feature.

PR #1166, opened 2026-07-04
Add `mp_show_hintmessages` cvar to allow blocking hint messages server-side

belomaxorka:
## What
Adds a new cvar `mp_show_hintmessages` (default `1`) that allows servers to disable sending hint messages (`#Hint_*`) to clients.

## Why
Server operators often want to get rid of the tutorial-style hint popups (e.g. `#Hint_you_have_the_bomb`, `#Hint_spotted_an_enemy`, zone hints). Currently this requires a plugin that hooks `RG_CBasePlayer_HintMessageEx` and supersedes it. Since every hint goes through a single function, this is trivial to support natively.

## How it works
All hint messages are routed through `CBasePlayer::HintMessageEx()` (both directly and via the `HintMessage()` wrapper) before being queued into `m_hintMessageQueue`. When `mp_show_hintmessages` is `0`, `HintMessageEx()` returns `false` immediately, so nothing is queued and nothing is sent to the client. The check is under `REGAMEDLL_ADD`; default value `1` keeps vanilla behavior.

## Edge cases
- Center-print messages (e.g. `#Got_bomb`) are not affected - they are not hints and bypass the hint queue by design.
- The API hookchain `RG_CBasePlayer_HintMessageEx` still fires as before; the cvar check is inside the hooked function body, so plugins keep working.
- The client-side "Auto-Help" setting (`m_bShowHints`) is untouched; the cvar acts as a server-wide override on top of it.

## Tested
Built Win32 Release (MSVC) and verified on a listen/dedicated ReHLDS server: with the cvar at `1` the bomb pickup hint is shown as usual; after `mp_show_hintmessages 0` (changeable at runtime) no hints are delivered, while center messages remain intact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_show_radioicon                  | 1       | 0   | 1            | Show radio icon.<br/>`0` disabled<br/>`1` enabled |
 | mp_show_scenarioicon               | 0       | 0   | 1            | Show scenario icon in HUD such as count of alive hostages or ticking bomb.<br/>`0` disabled<br/>`1` enabled |
 | mp_show_bomb_timer                 | 0       | 0   | 1            | Show the time until the planted bomb explodes on the HUD round timer.<br/>`0` disabled<br/>`1` enabled (when the bomb is planted, the round timer shows the C4 countdown) |
-| mp_show_hintmessages               | 1       | 0   | 1            | Send hint messages (Hint_*) to clients.<br/>`0` disabled (hints are blocked server-side)<br/>`1` enabled |
+| mp_show_hintmessages               | 1       | 0   | 1            | Send hint messages (Hint_*) to clients.<br/>`0` disabled (non-forced hints are suppressed server-side; forced hints, e.g. ReAPI `rg_hint_message` with override, still pass)<br/>`1` enabled |
 | mp_old_bomb_defused_sound          | 1       | 0   | 1            | Play "Bomb has been defused" sound instead of "Counter-Terrorists win" when bomb was defused<br/>`0` disabled<br/>`1` enabled |
 | showtriggers                       | 0       | 0   | 1            | Debug cvar shows triggers. |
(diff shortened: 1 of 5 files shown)


s1lentq:
ty for PR!
looks fine, but stuck in architectural deadlock here

issue is:
if cvar check stays inside `HintMessageEx` (like right now), it breaks `rg_hint_message` native for custom plugins
but if moving check outside directly to game code, it breaks `RG_CBasePlayer_HintMessageEx` hook for plugins that need to catch vanilla events
so no clean way to fix this on engine side without breaking ReAPI for someone

i am not sure, but maybe better to shift this responsibility to plugins
since impossible to solve all cases here, doing it via 3rd-party is much more flexible

belomaxorka:
> ty for PR! looks fine, but stuck in architectural deadlock here
> 
> issue is: if cvar check stays inside `HintMessageEx` (like right now), it breaks `rg_hint_message` native for custom plugins but if moving check outside directly to game code, it breaks `RG_CBasePlayer_HintMessageEx` hook for plugins that need to catch vanilla events so no clean way to fix this on engine side without breaking ReAPI for someone
> 
> i am not sure, but maybe better to shift this responsibility to plugins since impossible to solve all cases here, doing it via 3rd-party is much more flexible


Thanks for the detailed breakdown — you're right that there's no clean way to do vanilla-only suppression on the gamedll side. Both vanilla and plugin hints funnel through the same CBasePlayer::HintMessageEx, so there's no origin signal to key on, and gating at the call sites would hide vanilla events from RG_CBasePlayer_HintMessageEx hooks.

I've reframed the change rather than trying to solve the unsolvable case. The cvar is now positioned as a master kill-switch, not a vanilla-only filter:

- Default 1 — zero behavior change, vanilla builds untouched.
- 0 — suppresses non-forced hints server-side. The check sits in the HintMessageEx OrigFunc (chain terminal), so plugin hooks still fire and can observe every event.
- bOverride bypasses it, exactly as it already bypasses the per-client m_bShowHints (_ah) preference. So it doesn't introduce a new failure mode — non-override hints were already suppressible per client — but I'll be honest that it does widen the scope: with the cvar off, non-override
- rg_hint_message calls are dropped for all clients, so a plugin that needs guaranteed delivery must pass override.

I fully agree this doesn't cover every case, and forcing a hint against the server's wishes is reasonably a plugin's responsibility. So I'm leaving the call to you (@s1lentq) — if you feel this doesn't belong in the core and is better left to third-party plugins, please go ahead and close the PR. No hard feelings at all.

Issue #1125, opened 2025-10-14
[BUG]: Switching Condition Zero Hostage Issue


XM2323:
### 🤔 What happened?

During implementation in CS1.6, if Terrorists open fire, hostages may scatter (if the character locates the hostage position but the hostage is too far, it will snap back). Characters can get stuck with hostages (can only resolve by killing them; if the hostage is in an invincible state, they remain permanently stuck).

### ⚠️ Meta-information

```shell
dll edition：5.30.0.805

Game Version：48（1.1.2.6，4554）
```

### 📄 Relevant log output

```shell

```

Nord1cWarr1or:
CZ hostage behavior (scatter on gunfire, stuck in narrow passages) is Condition Zero specifics without repro or analysis — nothing actionable in the gamedll (cf. #724, #900, #979, #1059, and #1129 for the `hostage_ai_enable` gating). Closing as not planned.

Nord1cWarr1or:
Reopening as a valid candidate. The CZ-style hostage AI in CS 1.6 is a ReGameDLL feature enabled by `hostage_ai_enable` (default off) — it activates the improv AI port (`regamedll/dlls/hostage/hostage_improv.cpp` + state machine: idle/follow/retreat/escape) when the map has nav mesh data (`player.cpp:4335`). The reported behaviors map to that code: hostages scattering on gunfire is the retreat/escape-to-cover states (`regamedll/dlls/hostage/hostage_states.h`), and the "snap back"/stuck cases would be in the improv movement/localnav code. The report includes versions (dll 5.30.0.805, protocol 48), which helps.

To move this forward we need a concrete repro: map with a `.nav` file, `hostage_ai_enable 1`, and steps (hostage distance from player when firing starts, where exactly it gets stuck). Related: #1129 and the hostage crash family.

Nord1cWarr1or:
[...]

PR #1167, opened 2026-07-05
Add `mp_hostage_fear` cvar to control improved hostage AI fear reactions

belomaxorka:
Addresses the hostage panic/scatter part of #1125.

## What

New cvar `mp_hostage_fear` (default `1` - vanilla behavior) that controls whether hostages driven by the improved AI (`hostage_ai_enable 1`) get scared of gunfire, explosions and nearby deaths. With `mp_hostage_fear 0` hostages stay calm: they don't panic, don't scatter and don't refuse to follow after nearby shooting - CS 1.6-like hostage temperament while keeping the rest of the improved AI (navigation, chatter, animations).

## Why

With `hostage_ai_enable 1` on CS 1.6 hostage maps, any nearby enemy gunfire frightens the hostages (`CHostageImprov::OnGameEvent` -> `Frighten(SCARED/TERRIFIED)`). Scared hostages scatter, crouch and refuse to follow the rescuer, which #1125 reports as disruptive on 1.6 servers, where the original hostages ignore gunfire completely. Server operators currently have no way to keep the improved AI but disable the fear reactions.

## How it works

All fear-driven behavior (fleeing in `HostageIdleState`, refusing to follow, scared/afraid animations) is gated by `CHostageImprov::IsScared()`, which is true only while `m_scaredTimer` is running - and that timer is started exclusively in `CHostageImprov::Frighten()`. So a single early return in `Frighten()` when `mp_hostage_fear` is `0` keeps hostages calm everywhere. The cvar is evaluated on every call, so it can be flipped at runtime without a map change or server restart.

## Edge cases

- Direct damage to a hostage still makes it flee (`OnInjury` -> `m_mustFlee` in the idle state). That path is deliberately not gated: fleeing from an actual injury is self-preservation, distinct from panicking at noise.
- Classic 1.6 hostage AI (`hostage_ai_enable 0`) is unaffected - it has no fear logic at all.
- Default value `1` keeps vanilla behavior; vanilla builds are unaffected (the check is under `#ifdef REGAMEDLL_ADD`).

## Tested

Windows build (MSVC), verified on a ReHLDS server with `hostage_ai_enable 1`:
- `mp_hostage_fear 1` (default): hostages panic, scatter and refuse to follow when a Terrorist fires nearby - unchanged behavior.
- `mp_hostage_fear 0` (set at runtime, no restart): hostages stay calm under fire and keep following normally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ This means that plugins that do binary code analysis (Orpheu for example) probab
 | mp_timelimit                       | 0       | -   | -            | Period between map rotations.<br />`0` means no limit |
 | mp_forcerespawn                    | 0       | 0   | -            | Players will automatically respawn when killed.<br/>`0` disabled<br/>`>0.00001` time delay to respawn |
 | mp_hostage_hurtable                | 1       | 0   | 1            | The hostages can take damage.<br/>`0` disabled<br/>`1` from any team<br/>`2` only from `CT`<br/>`3` only from `T` |
+| mp_hostage_fear                    | 1       | 0   | 1            | The hostages with improved AI (`hostage_ai_enable 1`) get scared of gunfire, explosions and nearby deaths.<br/>`0` disabled (hostages stay calm)<br/>`1` enabled |
 | mp_show_radioicon                  | 1       | 0   | 1            | Show radio icon.<br/>`0` disabled<br/>`1` enabled |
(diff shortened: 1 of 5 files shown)


JUNESYNGOTOFLEX:
it's great to see how the functionality of controlling the behavior of both bots and hostages is expanding by adding new cvars. @s1lentq  in my opnion, it would be nice to divide them into different categories of configuration files (something like game_init.cfg)
having a separate branch of cvars for bots and  hostages.

PR #1190, opened 2026-09-07
cmake: remove duplicate -Wl,-z,notext flags introduced in #1177

blshkv:
PR #1177 added `-Wl,-z,notext` unconditionally before the first `set_target_properties` call, which is correct and sufficient. However, the same flag was also prepended to `LINK_FLAGS` inside each of the three `set_target_properties` branches (Android, non-Android Xash, non-Xash), causing it to be appended to `LINK_FLAGS` four times total.

This PR removes the three redundant copies, keeping only the single unconditional addition.


blshkv:
The Linux CI check is failing due to transient apt 404 errors on Debian Bullseye security mirrors (Bullseye is EOL), not related to this PR. Windows build passes fine.

